PR #34, opened 2023-09-01
Add extra check for byte count

rohitpaulk:


rohitpaulk:
Thanks to @alxkazinets for reporting this!

PR #56, opened 2024-02-11
Update Stage 3 (INFO replication on Replica) test.

ryan-gang:
We can't rely on users to make a connection to our master in this stage.
So, instead of blocking the main thread, we start a go routine to handle the connecting and performing an optional handshake.
Freeing our main thread to proceed as usual.

### DIFF
--- a/internal/stages_test.go
+++ b/internal/stages_test.go
@@ -239,7 +239,7 @@ func normalizeTesterOutput(testerOutput []byte) []byte {
 		"tcp_port":       {regexp.MustCompile("read tcp 127.0.0.1:\\d+->127.0.0.1:6379: read: connection reset by peer")},
 		" tmp_dir ":      {regexp.MustCompile(" /private/var/folders/[^ ]+ "), regexp.MustCompile(" /tmp/[^ ]+ ")},
 		"timestamp":      {regexp.MustCompile("\\d{2}:\\d{2}:\\d{2}\\.\\d{3}")},
-		"replication_id": {regexp.MustCompile("FULLRESYNC [A-Za-z0-9]+ 0 received.")},
+		"replication_id": {regexp.MustCompile(".*FULLRESYNC [A-Za-z0-9]+ 0 received.")},
 		"wait_timeout":   {regexp.MustCompile("WAIT command returned after [0-9]+ ms")},
 	}
 
--- a/internal/test_repl_info_replica.go
+++ b/internal/test_repl_info_replica.go
@@ -4,9 +4,9 @@ import (
 	"fmt"
 	"net"
 	"strings"
-	"time"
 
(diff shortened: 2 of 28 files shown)


internal/test_repl_info_replica.go
```diff
@@ -28,11 +29,22 @@ func testReplInfoReplica(stageHarness *testerutils.StageHarness) error {
 		return err
 	}
 
-	conn, err := listener.Accept()
-	if err != nil {
-		fmt.Println("Error accepting: ", err.Error())
-		return err
-	}
+	go func(l net.Listener) error {
```
rohitpaulk, line 32: Let's also remove the "Master is running on port 6379" log? I think it might be confusing to see that when one is only thinking of implementing INFO & `--replicaof`

internal/stages_test.go
```diff
@@ -239,7 +239,7 @@ func normalizeTesterOutput(testerOutput []byte) []byte {
 		"tcp_port":       {regexp.MustCompile("read tcp 127.0.0.1:\\d+->127.0.0.1:6379: read: connection reset by peer")},
 		" tmp_dir ":      {regexp.MustCompile(" /private/var/folders/[^ ]+ "), regexp.MustCompile(" /tmp/[^ ]+ ")},
 		"timestamp":      {regexp.MustCompile("\\d{2}:\\d{2}:\\d{2}\\.\\d{3}")},
-		"replication_id": {regexp.MustCompile("FULLRESYNC [A-Za-z0-9]+ 0 received.")},
+		"replication_id": {regexp.MustCompile(".*FULLRESYNC [A-Za-z0-9]+ 0 received.")},
```
rohitpaulk, line 242: @ryan-gang why do the characters before this need to be blanked out? I thought the only randomness was in the replication ID here?

rohitpaulk (approved):

internal/stages_test.go
ryan-gang: We have updated logging, now instead of the parsed RESP array, we are logging the raw RESP string we are sending, So just the + got added in the logs.

```
[stage-107] +PONG sent.
[stage-107] REPLCONF listening-port 6380 received.
[stage-107] +OK sent.
[stage-107] REPLCONF capa eof capa psync2 received.
[stage-107] +OK sent.
[stage-107] PSYNC ? -1 received.
[stage-107] +FULLRESYNC ig92fu64pmo8wp5vth6rx0nffooebbnvbc1h8qkn 0 sent.
```

rohitpaulk: That doesn’t explain the change to randomness - if you just regenerate fixtures with the old regex, things should work? 

This new regex ignores all content before FULLRESYNC, which can hide bugs

PR #57, opened 2024-02-12
Fixes to Redis Replication Extension tests.

ryan-gang:
Changelog : 

1. Revert regex change in stages_test.
2. Implement capability to compare user messages with multiple possible options.
(Used in capa stage, users would send `REPLCONF capa psync2`, REDIS would send `REPLCONF capa psync2 capa eof`. 
3. Some tests were not returning if connection establishment failed. Update those files to return right then.
4. Fixes to WAIT test. 
a. Return if WAIT reply mismatch.
b. Random numbers shouldn't be 0.


rohitpaulk:
@ryan-gang can we also remove the extra fixtures and just keep one per extension? (for the ones that use `pass_all`)

That'll make the diffs smaller + the tests will be faster on CI

ryan-gang:
Merging this PR, will create another PR with the changes you wanted @rohitpaulk.

PR #60, opened 2024-02-20
Fix offset tracking in WAIT test. 

ryan-gang:
## Changelog : 

- Fix offset tracking bug in `WAIT` test.
- Refactor `WAIT` test stage.
    Extract out `spawnReplicas`, `consumeReplicationStreamAndSendPartialAcks`, `RunWaitTest` methods.
- Break `util` file into `fake-redis-XYZ` helper files.
- Add helpful message to `INFO REPLICA` `bind error` err.
- Refactor bind error message, and add it to all `net.Listen()` calls.
- Update fixtures.


internal/test_repl_wait.go
```diff
@@ -2,15 +2,32 @@ package internal
...
+//     3.2. Issue a WAIT command with 1 as the expected number of replicas
+//     3.3. Read propagated command on replicas + respond to subset of GETACKs
+//     3.4. Assert response of WAIT command is 1
+//  4. Issue another write command, test WAIT <REPLICA_COUNT+1> 2000
+//     4.1. Issue a write command to the master
+//     4.2. Issue a WAIT command with a subset as the expected number of replicas
+//     4.3. Read propagated command on replicas + respond to subset of GETACKs
+//     4.4. Assert response of WAIT command is acks count
+//     4.5. Assert that the WAIT command returned after the timeout
 func testWait(stageHarness *testerutils.StageHarness) error {
```
rohitpaulk, line 27: @ryan-gang where is the refactored version of this? The diff here doesn't seem to include the extracted functions? 

Let's not repeat the same mistakes again and do this properly please

rohitpaulk (changes requested):
(Let's ship the cleanup along with this)

internal/test_repl_info_replica.go
```diff
@@ -10,14 +10,16 @@ import (
...
 
 func testReplInfoReplica(stageHarness *testerutils.StageHarness) error {
+	logger := stageHarness.Logger
+
 	listener, err := net.Listen("tcp", ":6379")
 	if err != nil {
-		fmt.Println("Error starting TCP server:", err)
+		logger.Errorf("Error starting TCP server: %v", err)
+		logFriendlyBindError(logger, err)
 		return err
```
rohitpaulk, line 19:
```suggestion
		return fmt.Errorf("Error starting TCP server: %v", err)
```

(if we do this and remove the logger.Efforf above, we won't have the error logged twice)

rohitpaulk (approved):

PR #62, opened 2024-02-22
CC-986 add support for streams - create stream

libmartinito:


internal/test_streams_xadd.go
```diff
@@ -0,0 +1,54 @@
...
+	if err := b.Run(); err != nil {
+		return err
+	}
+
+	logger := stageHarness.Logger
+
+	client := NewRedisClient("localhost:6379")
+
+	logger.Debugln("Sending XADD command")
+	logger.Infoln("$ redis-cli xadd stream_key 0-1 foo bar")
```
rohitpaulk, line 21: Let's use a random key please

rohitpaulk (approved):

internal/stages_test.go
```diff
@@ -76,7 +76,7 @@ func TestStages(t *testing.T) {
 			NormalizeOutputFunc: normalizeTesterOutput,
 		},
 		"streams_pass": {
-			UntilStageSlug:      "streams-type",
+			UntilStageSlug:      "streams-xadd",
 			CodePath:            "./test_helpers/pass_all",
 			ExpectedExitCode:    0,
 			StdoutFixturePath:   "./test_helpers/fixtures/streams-type/pass",
```
rohitpaulk, line 82: @libmartinito let's name this `fixtures/streams/pass`, and remove old fixtures as we go - currently I see a new `streams-xadd` fixtures added

internal/test_streams_xadd.go
```diff
@@ -0,0 +1,54 @@
...
+	b := NewRedisBinary(stageHarness)
+	if err := b.Run(); err != nil {
+		return err
+	}
+
+	logger := stageHarness.Logger
+
+	client := NewRedisClient("localhost:6379")
+
+	logger.Debugln("Sending XADD command")
```
rohitpaulk, line 20: This can probably be removed! Don't add much information really. 

An extra log here that might be informative as a debug log would be something like "Generated random stream key:  <the_random_value_generated>". That gives the user more insight into what the tester is doing.

PR #65, opened 2024-02-23
CC-986 add support for streams - query

libmartinito:


internal/test_streams_xrange.go
```diff
@@ -0,0 +1,88 @@
...
+			Values: map[string]interface{}{
+				"foo": "bar",
+			},
+		})
+	}
+
+	maxId := "0-" + strconv.Itoa(randomNumber)
+
+	logger.Infof("$ redis-cli xrange %s 0-1 %s", randomKey, maxId)
+	resp, err := client.XRange(randomKey, "0-1", maxId).Result()
```
rohitpaulk, line 76: @libmartinito I'd add two xrange tests - one that includes the full range, and one that only includes a subset. I think that users might naively implement xrange as "return everything" in the start and expect this test to fail

rohitpaulk (approved):
Needs success logs + another assertion around partial range, but otherwise looks good!

internal/test_streams_xrange.go
rohitpaulk: Actually, maybe one is sufficient - just that it shouldn't return the full set of entries and must only require a subset

PR #66, opened 2024-02-23
CC-986 add support for streams - query min

libmartinito:


rohitpaulk (approved):
(still needs success logging like the others)

PR #68, opened 2024-02-23
CC-986 add support for streams - validate-id

libmartinito:


internal/test_helpers/fixtures/streams/pass
```diff
@@ -512,3 +512,14 @@ Debug = true
...
 [33m[stage-33] [0m[36mTerminating program[0m
 [33m[stage-33] [0m[36mProgram terminated successfully[0m
+
+[33m[stage-34] [0m[94mRunning tests for Stage #34: streams-xadd-validate-id[0m
+[33m[stage-34] [0m[94m$ ./spawn_redis_server.sh[0m
+[33m[stage-34] [0m[94m$ redis-cli xadd mangos 1-1 foo bar[0m
+[33m[stage-34] [0m[94m$ redis-cli xadd mangos 1-2 bar baz[0m
+[33m[stage-34] [0m[94m$ redis-cli xadd mangos 1-2 baz foo[0m
+[33m[stage-34] [0m[94m$ redis-cli xadd mangos 0-3 baz foo[0m
+[33m[stage-34] [0m[94m$ redis-cli xadd mangos 0-0 baz foo[0m
```
rohitpaulk, line 522: Similar to other PRs - this needs success logs

internal/test_streams_xadd_validate_id.go
```diff
@@ -0,0 +1,121 @@
...
+		},
+	}).Result()
+
+	expectedErr := "ERR The ID specified in XADD is equal or smaller than the target stream top item"
+
+	if err.Error() != expectedErr {
+		return fmt.Errorf("Expected %#v, got %#v", expectedErr, err.Error())
+	}
+
+	logger.Infof("$ redis-cli xadd %s 0-3 baz foo", randomKey)
```
rohitpaulk, line 90: @libmartinito Lots of duplication across assertions within these tests + the other ID tests. I'd extract this into a utility function that can be re-used, that way we're sure that our logging is consistent across all tests.

rohitpaulk: Something like: 

```
type XADDTest { 
  streamKey string
  id string
  values map[string]string{"baz": "foo"}
  expectedResponse string // Either expectedResponse or expectedError will be set
  expectedError string
}
```

And then you could have a function that accepts that type along with logger etc. to run the test for you

rohitpaulk (dismissed):
Less duplication, ensure logs are consistent across all tests

internal/test_streams_xadd.go
```diff
@@ -5,9 +5,51 @@ import (
...
+		return err
+	}
+
+	logger.Infof("Received response: \"%s\"", resp)
+
+	if resp != test.expectedResponse {
+		return fmt.Errorf("Expected %#v, got %#v", test.expectedResponse, resp)
+	}
+
+	logger.Successf("Successfully added entry to stream")
```
rohitpaulk, line 48: Let's try to merge this with the log above? 

```
if (success) { 
  logger.Successf("Received response: XYZ")
} else {
  logger.Infof("Received response: XYZ")
  .. err contains details on expected vs. actual
}
```

(This avoids unnecessary verbosity in the success case)

rohitpaulk (approved):

PR #69, opened 2024-02-23
CC-986 add support for streams - partial-id

libmartinito:


internal/test_helpers/fixtures/streams/pass
```diff
@@ -512,3 +512,12 @@ Debug = true
 [33m[stage-33] [0m[92mTest passed.[0m
 [33m[stage-33] [0m[36mTerminating program[0m
 [33m[stage-33] [0m[36mProgram terminated successfully[0m
+
+[33m[stage-34] [0m[94mRunning tests for Stage #34: streams-xadd-partial-autoid[0m
+[33m[stage-34] [0m[94m$ ./spawn_redis_server.sh[0m
+[33m[stage-34] [0m[94m$ redis-cli xadd mangos 0-* foo bar[0m
+[33m[stage-34] [0m[94m$ redis-cli xadd mangos 1-* foo bar[0m
```
rohitpaulk, line 519: We need more success logs here @libmartinito. When users have errors, they'll want to know what their previous response were. Left notes on the full-id PR re: what this could look like.

rohitpaulk: The error logs look fine! So just adding 1 success log for each command should be good here I think

rohitpaulk (approved):

PR #70, opened 2024-02-23
CC-986 add support for streams - full-id

libmartinito:


internal/test_helpers/course_definition.yml
```diff
@@ -1482,3 +1482,41 @@ stages:
       - You can ignore the ability to parse multiple key value pairs.
     marketing_md: |
       In this stage, you'll add support for creating a [Redis stream](https://redis.io/docs/data-types/streams/) using the `XADD` command.
+
+  - slug: "streams-xadd-full-autoid"
```
rohitpaulk, line 1486: Was a bit confused by this - the next stage should be partialid, no?

internal/test_streams_xadd_full_autoid.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if err != nil {
+		logFriendlyError(logger, err)
+		return err
+	}
+
+	parts := strings.Split(resp, "-")
+
+	if len(parts) != 2 {
+		return fmt.Errorf("Expected 2 parts, got %d", len(parts))
```
rohitpaulk, line 55: At this point, since the user doesn't know that we're splitting by `-`, it might be confusing to see this error. 

For example, if I had this implemented as `0:1`, the error I'd see would be "Expected 2 parts, got 1". Can we be a more explicit and say "Expected a string of the form <integer>-<integer>, got XXX"?

internal/test_streams_xadd_full_autoid.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	time, sequenceNumber := parts[0], parts[1]
+
+	if len(time) != 13 {
+		return fmt.Errorf("Expected 13 characters, got %d", len(time))
```
rohitpaulk, line 61: Against, let's be more explicit here - "Expected the first part of the ID to be a unix timestamp (13 characters), got XXX". Remember that the user isn't reading your test code, the only feedback they have is through error messages.

rohitpaulk: We should also check the time here and make sure it is an actual unix timestamp (can add +/- 500ms threshold to avoid the exact check)

rohitpaulk: I know the number 13 won't change for a long time, but since we need to generate one for comparisons anyway, I'd recommend checking `len` based on the unix milliseconds integer that we generate, instead of hardcoding 13

internal/test_streams_xadd_full_autoid.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	if sequenceNumber != "0" {
+		return fmt.Errorf("Expected \"0\", got %#v", sequenceNumber)
```
rohitpaulk, line 65: Let's make it more clear what we're expecting here. "Expected the second part (sequence number) to be 0, got XYZ".

internal/test_streams_xadd_full_autoid.go
```diff
@@ -0,0 +1,69 @@
...
+			"foo": "bar",
+		},
+	}).Result()
+
+	if err != nil {
+		logFriendlyError(logger, err)
+		return err
+	}
+
+	parts := strings.Split(resp, "-")
```
rohitpaulk, line 52: Let's log out the `resp` with an INFO log here, and add two success logs, one for each part. 

Example logs that I think would be a good experience:

```
[INFO] $ redis-cli xadd ...
[INFO] Received response: "0-1".
[ERROR] Expected the first part of the ID to be a unix milliseconds timestamp (13 characters), got "0".
```


```
[INFO] $ redis-cli xadd ...
[INFO] Received response: "1708720384951-1".
[SUCCESS] First part of ID is a unix milliseconds timestamp.
[ERROR] Expected the second part of the ID (sequence number) to be 0, got "1".
```

rohitpaulk (approved):

PR #71, opened 2024-02-26
CC-1032: Improve anti-cheat stage.

ryan-gang:
Instead of `INFO Server`, use `MEMORY DOCTOR`.
Fixes https://github.com/codecrafters-io/build-your-own-redis/issues/127.

### DIFF
--- a/internal/anti_cheat.go
+++ b/internal/anti_cheat.go
@@ -23,7 +23,7 @@ func antiCheatTest(stageHarness *testerutils.StageHarness) error {
 	}
 	defer conn.Close()
 
-	if err := conn.SetReadDeadline(time.Now().Add(2 * time.Second)); err != nil {
+	if err := conn.SetReadDeadline(time.Now().Add(100 * time.Millisecond)); err != nil {
 		return fmt.Errorf("Error setting read deadline: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


ryan-gang:
Ref : https://github.com/redis/redis/blob/bfcaa7db0a70f81c062e2ff6dee1b23baf1cbf53/src/object.c#L1361

internal/anti_cheat.go
```diff
@@ -13,24 +15,49 @@ func antiCheatTest(stageHarness *testerutils.StageHarness) error {
...
+	if err != nil {
+		logger.Debugf("Error connecting to TCP server: %v", err)
+		return err
+	}
+	defer conn.Close()
 
-	result := client.Info("server")
-	if result.Err() != nil {
-		return nil
+	if err := conn.SetReadDeadline(time.Now().Add(2 * time.Second)); err != nil {
```
rohitpaulk, line 26: Can be even shorter here - let's do 100ms?

rohitpaulk: (Connect deadline needs to be reasonable, but read deadline can be super short)

internal/anti_cheat.go
```diff
@@ -13,24 +15,49 @@ func antiCheatTest(stageHarness *testerutils.StageHarness) error {
...
+		logger.Criticalf("Are you sure you aren't running this against the actual Redis?")
+		return fmt.Errorf("anti-cheat (ac1) failed")
 	}
+	return nil
+}
 
-	logger.Criticalf("anti-cheat (ac1) failed.")
-	logger.Criticalf("Are you sure you aren't running this against the actual Redis?")
-	return fmt.Errorf("anti-cheat (ac1) failed")
+func messageDetected(actualMessage string) bool {
```
rohitpaulk, line 51: Bad naming - name functions based on what they do. In this case, I'd suggest `validateMessage` / `checkMessage` / `detectMessage`

internal/anti_cheat.go
```diff
@@ -13,24 +15,49 @@ func antiCheatTest(stageHarness *testerutils.StageHarness) error {
...
+	var possibleMessages = []string{
+		"Hi Sam, I can't find any memory issue in your instance.",
```
rohitpaulk, line 53: Can this check be simpler? What about: 

- The response is a string response (not an error)
- The response contains the string "memory" in it

rohitpaulk (approved):

PR #72, opened 2024-02-26
CC-986 add support for streams - xread

libmartinito:


internal/test_helpers/fixtures/streams/pass
```diff
@@ -537,3 +537,13 @@ Debug = true
 [33m[stage-35] [0m[92mTest passed.[0m
 [33m[stage-35] [0m[36mTerminating program[0m
 [33m[stage-35] [0m[36mProgram terminated successfully[0m
+
+[33m[stage-36] [0m[94mRunning tests for Stage #36: streams-xread[0m
+[33m[stage-36] [0m[94m$ ./spawn_redis_server.sh[0m
+[33m[stage-36] [0m[94m$ redis-cli xadd pears 0-1 map[foo:bar][0m
```
rohitpaulk, line 543:
```suggestion
[33m[stage-36] [0m[94m$ redis-cli xadd pears 0-1 map[foo:bar][0m
```

@libmartinito why does this say "map[foo:bar]" here instead of just `xadd pears 0-1 foo bar`? Isn't that what the valid redis-cli command would be?

internal/test_helpers/fixtures/streams/pass
```diff
@@ -537,3 +537,13 @@ Debug = true
...
+[33m[stage-36] [0m[92mReceived response: "0-1"[0m
+[33m[stage-36] [0m[94m$ redis-cli xread streams pears 0-0[0m
+[33m[stage-36] [0m[92mReceived response: "[{pears [{0-1 map[foo:bar]}]}]"[0m
```
rohitpaulk, line 546: Let's print this as json? That's a more familiar format for users.

internal/test_streams_xread.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+type XREADTest struct {
+	streams          []string
+	block            *int
+	expectedResponse []redis.XStream
+	expectedError    error
+}
+
+func testXread(client *redis.Client, logger *logger.Logger, test XREADTest) error {
```
rohitpaulk, line 21: Bad naming - looking at the names, it isn't clear how "testStreamsXread` and `testXread` are different. I'd suggest naming this `executeXreadTest` or something like that

rohitpaulk: You could even make it a function on the type: 

```
func (t XReadTest) Run(client *redis.Client, logger *logger.Logger) error {}
```

rohitpaulk (approved):

PR #73, opened 2024-02-26
CC-986 add support for streams - xread-multiple

libmartinito:


internal/test_streams_xread_multiple.go
```diff
@@ -0,0 +1,90 @@
...
+	b := NewRedisBinary(stageHarness)
+	if err := b.Run(); err != nil {
+		return err
+	}
+
+	logger := stageHarness.Logger
+
+	client := NewRedisClient("localhost:6379")
+
+	strings := [10]string{
```
rohitpaulk, line 20: I've been seeing this repeated in multiple places - we already have utilities for this here btw: https://github.com/codecrafters-io/tester-utils/blob/master/random/helpers.go. Let's try to keep these tests short and to the point - someday they'll be visible within the app.

internal/test_streams_xread_multiple.go
```diff
@@ -0,0 +1,90 @@
...
+		"plums",
+	}
+
+	randomKey := strings[rand.Intn(10)]
+	otherRandomKey := otherStrings[rand.Intn(10)]
+
+	testXadd(client, logger, XADDTest{
+		streamKey:        randomKey,
+		id:               "0-1",
+		values:           map[string]interface{}{"foo": "bar"},
```
rohitpaulk, line 52: When testing `xread`, I think you'll need the values to be random too - seems super easy to hardcode stuff here and accidentally pass the stage. One easy way here could be to use fixed keys (temperature, humidity etc.) and use random integer values.

rohitpaulk (approved):

PR #75, opened 2024-02-27
CC-986 add support for streams - xread-block-max-id

libmartinito:


internal/test_helpers/fixtures/streams/pass
```diff
@@ -583,3 +583,14 @@ Debug = true
 [33m[stage-39] [0m[92mTest passed.[0m
 [33m[stage-39] [0m[36mTerminating program[0m
 [33m[stage-39] [0m[36mProgram terminated successfully[0m
+
+[33m[stage-40] [0m[94mRunning tests for Stage #40: streams-xread-block-max-id[0m
```
rohitpaulk, line 587: @libmartinito the XREAD command isn't logged here at all?

rohitpaulk (changes requested):
Added comment

PR #78, opened 2024-02-28
CC-986 add support for blocking xread

libmartinito:


internal/test_helpers/fixtures/streams/pass
```diff
@@ -651,3 +651,77 @@ Debug = true
...
+[33m[stage-42] [0m[92m        "ID": "0-2",[0m
+[33m[stage-42] [0m[92m        "Values": {[0m
+[33m[stage-42] [0m[92m          "temperature": "73"[0m
+[33m[stage-42] [0m[92m        }[0m
+[33m[stage-42] [0m[92m      }[0m
+[33m[stage-42] [0m[92m    ][0m
+[33m[stage-42] [0m[92m  }[0m
+[33m[stage-42] [0m[92m]"[0m
+[33m[stage-42] [0m[94m$ redis-cli xread block 1000 streams pineapple 0-2[0m
+[33m[stage-42] [0m[92mReceived error: "redis: nil"[0m
```
libmartinito, line 676: not sure if we should remove the part around null for this stage? it seems that xread returns an error when blocking times out and nothing is received, but this doesn't appear in the docs.

rohitpaulk: @libmartinito This shoudn't be reported as an error - we should just say "Received nil" (and we should _expect_ that as the reply too). Check the expiry stage for how this is handled - there too we expect a nil response.

internal/test_helpers/fixtures/streams/pass
```diff
@@ -651,3 +651,77 @@ Debug = true
...
+[33m[stage-43] [0m[94mRunning tests for Stage #43: streams-xread-block-no-timeout[0m
+[33m[stage-43] [0m[94m$ ./spawn_redis_server.sh[0m
+[33m[stage-43] [0m[94m$ redis-cli xadd raspberry 0-1 temperature 54[0m
+[33m[stage-43] [0m[92mReceived response: "0-1"[0m
+[33m[stage-43] [0m[94m$ redis-cli xread block 0 streams raspberry 0-1[0m
+[33m[stage-43] [0m[94m$ redis-cli xadd raspberry 0-2 temperature 54[0m
+[33m[stage-43] [0m[92mReceived response: "0-2"[0m
+[33m[stage-43] [0m[92mReceived response: "[[0m
+[33m[stage-43] [0m[92m  {[0m
+[33m[stage-43] [0m[92m    "Stream": "raspberry",[0m
```
rohitpaulk, line 690: Are these actual keys returned by the user's response? 🤔 I thought this is just a flat array like `["raspberry", ...]`? 

We shouldn't add extra keys to the response if we're printing it out for debugging purposes. JSON formatting is fine, but the structure of the message should stay intact

rohitpaulk (approved):

internal/test_helpers/fixtures/streams/pass
libmartinito: will revisit after actually going through the extension

PR #79, opened 2024-02-28
Add resp modules, use in base stages

rohitpaulk:


internal/resp/client/client.go
```diff
@@ -0,0 +1,192 @@
...
+		_, _, err := resp_decoder.Decode(c.UnreadBuffer.Bytes())
+
+		if err == nil {
+			break // We were able to read a value!
+		}
+
+		if _, ok := err.(resp_decoder.InvalidRESPError); ok {
+			break // We've read an invalid value, we can stop reading immediately
+		}
+	}
```
ryan-gang, line 144: As there is no explicit wait period here, if the user were to serve "wrong" reads (so that ReadDeadline was skipped), wouldn't we hammer their service ? 
(In some specific cases of IncompleteRESPError).

rohitpaulk: I improved this a bit later: 

Screenshot 2024-03-02 at 12 28 41

In reality we aren't "hammering" their service since this is just a TCP read - we're just calling a syscall far too many times, the user's program doesn't even see this call (unlike a TCP connect, which the program will see). So as long as we don't end up in a busy loop, we should be fine - the short sleep fixes that. 

(@ryan-gang let me know if you were thinking of another case?)

ryan-gang: No, I feel the updated one is much better. 
I was initially thinking that we would be calling the condition function every time we read a byte, but we are reading in chunks of 1024 bytes, so we are good on that front too.

PR #83, opened 2024-02-29
Use new resp parser in one replication stage

rohitpaulk:
This adds all the functionality needed for the replication propagation stage to the new RESP parser & wires it up in that stage alone. Let's handle other stages in separate PRs, this is pretty huge already


rohitpaulk:
@ryan-gang can you review this? In particular, look at whether there's anything wrong with the replication test logic + whether the structure is extensible enough for you to go ahead and wire this up in all other replication stages (sometime next week).

I tried to cover all the "difficult" cases I could think of - like skipping SELECT, case-insensitive matching etc.

internal/resp/decoder/errors.go
```diff
@@ -8,21 +8,21 @@ import (
 	inspectable_byte_string "github.com/codecrafters-io/redis-tester/internal/inspectable_byte_string"
 )
 
-type IncompleteRESPError struct {
+type IncompleteInputError struct {
```
ryan-gang, line 11: The previous error names made more sense in this context, no ?

ryan-gang:
Currently `(resp_assertions.RegexStringAssertion).Run` is case sensitive. To handle case insensitive matches, should we use `(resp_assertions.RegexStringAssertion).Run` ?

ryan-gang:
We would need to add another `CommandTestCase` like struct, where we read the user's command first, assert and then send the reply. 
To be used when we act as master, and respond to user's replica asking to sync.

internal/test_repl_master_cmd_prop.go
```diff
@@ -1,42 +1,46 @@
...
 	}
 
-	conn1, err := NewRedisConn("", "localhost:6379")
+	// We use another client to assert whether sent commands are replicated from the master (user's code)
+	replicaClient, err := instrumented_resp_client.NewInstrumentedRespClient(stageHarness, "localhost:6379", "replica")
 	if err != nil {
-		fmt.Println("Error connecting to TCP server:", err)
+		logFriendlyError(logger, err)
 		return err
 	}
```
ryan-gang, line 41:
```suggestion
	defer replicaClient.Close()
```

internal/test_repl_master_cmd_prop.go
```diff
@@ -1,42 +1,46 @@
...
 	logger := stageHarness.Logger
 
-	conn, err := NewRedisConn("", "localhost:6379")
+	// We use one client to send commands to the master
+	client, err := instrumented_resp_client.NewInstrumentedRespClient(stageHarness, "localhost:6379", "client")
 	if err != nil {
-		fmt.Println("Error connecting to TCP server:", err)
+		logFriendlyError(logger, err)
 		return err
 	}
```
ryan-gang, line 34:
```suggestion
	defer client.Close()
```

internal/test_repl_master_cmd_prop.go
ryan-gang:
```
[stage-111] Test passed.
[stage-111] Terminating program
[stage-111] Error terminating program: 'program failed to exit in 2 seconds after receiving sigterm'
```

ryan-gang (approved):
LGTM!

internal/test_repl_master_cmd_prop.go
rohitpaulk: Ah, so that was the reason! Nice catch. I thought that in replication mode Redis does some kind of clean shutdown procedure, this makes more sense

internal/resp/decoder/errors.go
rohitpaulk: I changed this because of the DecodeRDBFile method. Feels weird for DecodeRDBFile to return InvalidRESP

rohitpaulk: If we have more "exceptions" to resp like this, then might make sense to move to a separate package or so - since it's just one for now, don't think it's a big deal

rohitpaulk:
> Currently (resp_assertions.RegexStringAssertion).Run is case sensitive. To handle case insensitive matches, should we use (resp_assertions.RegexStringAssertion).Run ?

Good question. Depends on the context - you're right that we _can_ pass a case-insensitive regex to `RegexStringAssertion`. The resulting error message won't be great though. 

With assertions, I think we should optimize for the error message first, and re-usability next.

For commands, the case-insensitivity is handled implicitly within the `CommandAssertion.Run` logic, and the error message always returns the expected & actual in lowercase, so it shouldn't be confusing. 

For other cases, we should look at this similarly - how can we emit a non-confusing error message? And then work our way backwards to see which assertions might work, and whether we need new ones. 

@ryan-gang btw did you have a specific case in mind for this question? I thought that commands were the only case where we have a need for case-insensitive matching?

rohitpaulk:
> We would need to add another CommandTestCase like struct, where we read the user's command first, assert and then send the reply.
> To be used when we act as master, and respond to user's replica asking to sync.

Cool, yeah good point. I did expect that we'd need to create a `SendReplicationHandshakeTestCase`, but didn't think of the fact that it wouldn't be able to re-use `CommandTestCase`. Maybe we name rename `CommandTestCase` to  `SendCommandTestCase` and then add `ReceiveCommandTestCase` for the case you're mentioning?

internal/resp/decoder/errors.go
ryan-gang: Makes sense.

ryan-gang:
> @ryan-gang btw did you have a specific case in mind for this question? I thought that commands were the only case where we have a need for case-insensitive matching?

Nothing comes to mind right away, we should be set with the CommandAssertion thing for now. I was inquiring more from an extensibility point of view.

PR #94, opened 2024-03-05
CC-1063: Update test to match stage description.

ryan-gang:


internal/test_repl_replica_getack_nonzero.go
```diff
@@ -51,6 +51,12 @@ func testReplGetaAckNonZero(stageHarness *test_case_harness.TestCaseHarness) err
 	}
 	offset += GetByteOffset([]string{"REPLCONF", "GETACK", "*"})
 
+	err = master.GetAck(offset) // 37
```
rohitpaulk, line 54: @ryan-gang how come the values in the comments above (line 48) and below (line 64) is also `37`?

rohitpaulk (changes requested):
(comments added)

internal/test_repl_replica_getack_nonzero.go
ryan-gang: These are the byte offsets for the given command. Not the cumulative offset value. These were useful during writing the initial tests.
Should I remove these comments altogether ?

rohitpaulk: Aha, no it's okay then

rohitpaulk (approved):

qwertzguy:
This technically still doesn't match the test description as the description mentions only sending one SET command but the test sends 2 of them.

ryan-gang:
We want 2 set of `SET` commands, it makes the tests more robust. Will update the description to mention the same.

PR #98, opened 2024-03-06
fix the incorrect logic for checking the return value of WAIT command

youngk1019:
fix the incorrect logic for checking the return value of WAIT command
The problem comes from https://github.com/codecrafters-io/build-your-own-redis/issues/141

### DIFF
--- a/internal/fake_redis_master.go
+++ b/internal/fake_redis_master.go
@@ -1,9 +1,11 @@
 package internal
 
 import (
+	"fmt"
 	"net"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/codecrafters-io/tester-utils/logger"
 )
@@ -66,8 +68,26 @@ func (master FakeRedisMaster) GetAck(offset int) error {
 	return master.SendAndAssertStringArray([]string{"REPLCONF", "GETACK", "*"}, []string{"REPLCONF", "ACK", strconv.Itoa(offset)})
 }
 
-func (master FakeRedisMaster) Wait(replicas string, timeout string, expectedMessage int) error {
-	return master.SendAndAssertInt([]string{"WAIT", replicas, timeout}, expectedMessage)
+func (master FakeRedisMaster) Wait(needReplicas int, timeout time.Duration, maxReplicas int) error {
+	if err := master.Send([]string{"WAIT", strconv.Itoa(needReplicas), strconv.Itoa(int(timeout.Milliseconds()))}); err != nil {
+		return err
+	}
+	actualMessage, err := master.readRespInt()
+	if err != nil {
+		return err
+	}
+
+	if needReplicas > maxReplicas {
+		if actualMessage != maxReplicas {
+			return fmt.Errorf("Expected: '%v' and actual: '%v' messages don't match", maxReplicas, actualMessage)
+		}
+	} else {
+		if actualMessage < needReplicas || actualMessage > maxReplicas {
+			return fmt.Errorf("Expected: '[%v, %v]' and actual: '%v' messages don't match", needReplicas, maxReplicas, actualMessage)
+		}
+	}
+	master.Logger.Successf(master.LogPrefix + strconv.Itoa(actualMessage) + " received.")
+	return nil
 }
 
 func (master FakeRedisMaster) Handshake() error {
--- a/internal/test_repl_wait_zero_offset.go
+++ b/internal/test_repl_wait_zero_offset.go
@@ -2,10 +2,9 @@ package internal
 
 import (
 	"fmt"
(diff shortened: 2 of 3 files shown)


internal/fake_redis_master.go
```diff
@@ -66,8 +68,26 @@ func (master FakeRedisMaster) GetAck(offset int) error {
...
+	if err != nil {
+		return err
+	}
+
+	if needReplicas > maxReplicas {
+		if actualMessage != maxReplicas {
+			return fmt.Errorf("Expected: '%v' and actual: '%v' messages don't match", maxReplicas, actualMessage)
+		}
+	} else {
+		if actualMessage < needReplicas || actualMessage > maxReplicas {
```
qwertzguy, line 85: The previous logic was correct in this test since the offset is zero, and it waits for all replicas to finish the handshake, all replicas have reached the zero offset. In the Redis spec you can see that the retrun value for WAIT is the total number of replicas  that have reached the offset. So it would be incorrect to return a value less than maxReplicas.

qwertzguy (changes requested):
(FYI I'm not part of codecrafters) I think this change would let incorrect implementations pass the tests.

internal/fake_redis_master.go
```diff
@@ -66,8 +68,26 @@ func (master FakeRedisMaster) GetAck(offset int) error {
 	return master.SendAndAssertStringArray([]string{"REPLCONF", "GETACK", "*"}, []string{"REPLCONF", "ACK", strconv.Itoa(offset)})
 }
 
-func (master FakeRedisMaster) Wait(replicas string, timeout string, expectedMessage int) error {
-	return master.SendAndAssertInt([]string{"WAIT", replicas, timeout}, expectedMessage)
+func (master FakeRedisMaster) Wait(needReplicas int, timeout time.Duration, maxReplicas int) error {
```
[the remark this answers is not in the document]
youngk1019: Originally I wanted to check the timeout, so I modified the parameter type, but later I found that if I needed to check the timeout, the logic would become very complicated, and a lot of content would need to be added like the next test.

internal/fake_redis_master.go
qwertzguy: I realized I forgot to paste the link to the Redis spec: https://redis.io/commands/wait/#resp3-reply.
From the spec, my understanding is that it should return the total number of replicas that match or exceed the offset. I understand that it's a special case with the offset zero, but for the purpose of this test I think it should assert for the total replicas, as someone would need to have some weird logic not to return that, and this logic complicates this code. Going to remove my request changes to unblock though as it wouldn't really break the tests, even though it makes it more lenient  than it should be.

qwertzguy: Seems like I can't remove the requested changes without approving, but I actually think since I'm not a maintainer it doesn't block merging.

youngk1019: I can understand what you mean, you think that for example all replicas have completed synchronization. However, I think that forcing the return of the maximum number of replicas is not a correct way to check https://redis.io/commands/wait/ , because I think the way to check the offset for each replica is asynchronous. There may be unsynchronized commands in the current replica before the wait command. I need Waiting for the synchronization of the previous command to perform the wait check, then when I receive enough replicas to tell me that the requirements have been met, I can return. This is the return logic that complies with the rules. If in order to achieve the current requirements, On the contrary, it will degrade the performance of my design, because I need to find a centralized place to maintain the information of all replicas, so that I can first check synchronously whether the current replica has been synchronized, which will greatly increase the situation of lock competition.

qwertzguy: Yes this test is definitely a special case, which is that no commands have been sent, so by default any replica that connects is already up-to-date. If this was testing an offset other than zero, than I would agree with the need for the new logic you wrote.
As for your implementation, I think that before locking to wait for replicas it should first check which replicas are behind the required offset, and only getack on those, not the ones where you already know they are up-to-date from a previous async getack. Simply checking which replicas are behind the offset should be very fast, so you can either do it synchronously (it would probably be even faster since there would be less context switch and more contiguous memory reads from the CPU) or do it async and wait for all of them to finish (which again should be almost instant since it's just an int comp)

youngk1019: In my implementation this will not be faster, the faster situation is only in this special case, because multiple threads are required to synchronize data in the same address space.
And I think this is not the focus of the discussion. The point is that this checking method does not comply with the specifications given by redis. It does not matter how individuals implement it.

rohitpaulk:
Note added here: https://github.com/codecrafters-io/build-your-own-redis/issues/141#issuecomment-2019015521

PR #99, opened 2024-03-07
Fix tests for "WAIT with multiple commands" stage: correct ACK offset + respond to all GETACKs

qwertzguy:
In the tests for the "WAIT with multiple commands" stage, the responses to the GETACK command from the master would have the incorrect offset: the offset would include the bytes of the REPLCONF GETACK command itself but it shouldn't, those bytes should only be included in the next offset.
Also, for masters that would handle async GETACKs and track which GETACKs are in progress in order to not double send a GETACK, a subset of replicas would ignore the first GETACK command and never reply to it, making the master continue to wait indefinitely for those as a previous GETACK was still pending.
This PR fixes both of those issues by:
- Fixing the ACK offset returned from replicas by not including the current GETACK command
- Have the remainder replicas respond to GETACKs after the WAIT command was asserted in order to not change the test

FYI: I have not tested this change.


rohitpaulk:
I believe this is fixed now! We did push some changes to the ACK offset stages + updated the instructions to match the tester too.

PR #104, opened 2024-03-11
CC-1059: Use new Redis parser & assertions in Replication extension.

ryan-gang:


internal/anti_cheat.go
```diff
@@ -2,46 +2,41 @@ package internal
...
+	"github.com/codecrafters-io/redis-tester/internal/test_cases"
 	"github.com/codecrafters-io/tester-utils/test_case_harness"
 )
 
 func antiCheatTest(stageHarness *test_case_harness.TestCaseHarness) error {
-	b := NewRedisBinary(stageHarness)
-	if err := b.Run(); err != nil {
+	b := redis_executable.NewRedisExecutable(stageHarness)
+	if err := b.Run([]string{
+		"--port", "6379",
```
rohitpaulk, line 16: @ryan-gang is the argument required here? Earlier we were just running this without any args.

internal/anti_cheat.go
```diff
@@ -2,46 +2,41 @@ package internal
...
-	if err != nil {
-		if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
-			return nil // Read timed out. No data received from client.
-		}
-		return nil
+	// All the answers for MEMORY DOCTOR include the string "sam" in them.
+	commandTestCase := test_cases.CommandTestCase{
+		Command:                   "MEMORY",
+		Args:                      []string{"DOCTOR"},
+		Assertion:                 resp_assertions.NewRegexStringAssertion("[\\s\\S]*[sS]am[\\s\\S]*"),
```
rohitpaulk, line 34: Is there not a simpler form for this? Would just `[sS]am` work?

internal/instrumented_resp_connection/instrumented_resp_connection.go
```diff
@@ -0,0 +1,47 @@
...
+	logPrefix := ""
+	if clientIdentifier != "" {
+		logPrefix = clientIdentifier + ": "
+	}
+	return resp_connection.NewRespClientWithCallbacks(
+		addr, defaultCallbacks(stageHarness, logPrefix),
+	)
+}
+
+func NewInstrumentedRespConnection(stageHarness *test_case_harness.TestCaseHarness, conn net.Conn, clientIdentifier string) (*resp_connection.RespConnection, error) {
```
rohitpaulk, line 38: The names here are weird. Both `NewInstrumentedRespClient` and `NewInstrumentedRespConnection` return a `RespConnection` - from the names you can't tell that the difference is that they accept different params. 

If the primary difference here is in the arguments, we should name it something like `NewInstrumentedRespConnectionFromAddr`, `NewInstrumentedRespConnectionFromConn`. 

If the whole package only exports one type, you can even shorten this and make it `NewFromAddr`, `NewFromConn`.

rohitpaulk: I haven't looked at the differences in usage in detail, but I suspect we'd want the logging style to be different for a standard client vs. one that's used for the replication stream. Doesn't make sense to say `$ redis-cli PING` when we're sending a command via the replication stream, the logs would be misleading

internal/redis_executable/redis_executable_helper.go
rohitpaulk, line 1: The filename can just be `redis_executable.go`, `_helper` doesn't add any value here

internal/test_helpers/fixtures/repl-wait/pass
```diff
@@ -11,7 +11,7 @@ Debug = true
 [33m[stage-2] [0m[94mRunning tests for Stage #2: ping-pong[0m
 [33m[stage-2] [0m[94m$ ./spawn_redis_server.sh[0m
 [33m[stage-2] [0m[36mConnection established, sending ping command...[0m
-[33m[stage-2] [0m[94m$ redis-cli ping[0m
+[33m[stage-2] [0m[94m$ redis-cli ["ping"][0m
```
rohitpaulk, line 14: @ryan-gang this is clearly wrong, why open something for review when there's an obvious bug like this?

rohitpaulk (changes requested):
Split up - there's too much wrong here, let's review independent chunks

PR #105, opened 2024-03-14
CC-1059: implement builtin offset tracking to `respConnection`

ryan-gang:


internal/resp/connection/connection.go
```diff
@@ -42,6 +42,13 @@ type RespConnection struct {
 
 	// Callbacks is a set of functions that are called at various points in the client's lifecycle.
 	Callbacks RespConnectionCallbacks
+
+	// Offset tracking:
+	// SentBytes is the number of bytes sent to the server.
```
rohitpaulk, line 47:
```suggestion
	// SentBytes is the number of bytes sent using this connection.
```

We don't specify "server vs. client" anywhere in this file

internal/resp/connection/connection.go
```diff
@@ -42,6 +42,13 @@ type RespConnection struct {
...
+	SentBytes int
+
+	// ReceivedBytes is the number of bytes received from the server.
```
rohitpaulk, line 50:
```suggestion
	// ReceivedBytes is the number of bytes received using this connection
```

internal/resp/connection/connection.go
```diff
@@ -89,7 +96,11 @@ func (c *RespConnection) SendCommand(command resp_value.Value) error {
 	}
 
 	encodedValue := resp_encoder.Encode(command)
-	return c.SendBytes(encodedValue)
+	if err := c.SendBytes(encodedValue); err != nil {
+		return err
+	}
+	c.SentBytes += len(encodedValue)
```
rohitpaulk, line 102: Why aren't we incrementing this in the `SendBytes` function?

internal/resp/connection/connection.go
```diff
@@ -217,6 +228,13 @@ func (c *RespConnection) readIntoBufferUntil(condition func([]byte) bool, timeou
 	}
 }
 
+func (c *RespConnection) InitializeOffset() {
```
rohitpaulk, line 231: Why are we using the name `Offset` here? It doesn't mean anything in the context of a `RespConnection`. 

Better names: 

- `ResetReceivedAndSentBytesCount()`
- `ResetByteCounters()`

PR #106, opened 2024-03-14
CC-1059: 1. Replace NewRedisBinary with redis_executable.NewRedisExecutable.

ryan-gang:
Update usage in all test files.

### DIFF
--- a/internal/test_rdb_config.go
+++ b/internal/test_rdb_config.go
@@ -2,6 +2,7 @@ package internal
 
 import (
 	"fmt"
+	"github.com/codecrafters-io/redis-tester/internal/redis_executable"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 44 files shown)


internal/test_rdb_config.go
```diff
@@ -23,13 +24,11 @@ func testRdbConfig(stageHarness *test_case_harness.TestCaseHarness) error {
 	}
 	tmpDir = realPath
 
-	b := NewRedisBinary(stageHarness)
-	b.args = []string{
+	b := redis_executable.NewRedisExecutable(stageHarness)
+	if err := b.Run([]string{
```
rohitpaulk, line 28: Can just pass strings directly, since it's variadic? 

`b.Run(a, b, c, d)`

rohitpaulk (approved):

PR #107, opened 2024-03-15
CC-1059: 2. Refactor `bind_test_case`.

ryan-gang:
Refactor test cases to use a common BindTestCase struct and method for testing connection to ports.

### DIFF
--- a/internal/test_bind.go
+++ b/internal/test_bind.go
@@ -7,8 +7,6 @@ import (
 )
 
 func testBindToPort(stageHarness *test_case_harness.TestCaseHarness) error {
-	port := 6379
-
 	b := redis_executable.NewRedisExecutable(stageHarness)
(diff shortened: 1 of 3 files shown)


internal/test_bind.go
```diff
@@ -1,50 +1,27 @@
...
-		if err != nil {
-			if b.HasExited() {
-				return fmt.Errorf("Looks like your program has terminated. A redis server is expected to be a long-running process.")
-			}
-
-			// Don't print errors in the first second
-			if retries > 2 {
-				logger.Infof("Failed to connect to port 6379, retrying in 1s")
-			}
+	logger.Infof("Connecting to port %d...", port)
```
rohitpaulk, line 19: This can probably be moved into bind_test_case.go

rohitpaulk (approved):

PR #109, opened 2024-03-15
CC-1059: Refactor command propagation stages.

ryan-gang:


rohitpaulk:
(dismissing review since this needs to be rebased / cleaned up)

Makefile
```diff
@@ -24,9 +24,9 @@ test_with_redis: build
 
 
 test_tmp: build
-	cd /tmp/abc && \
-	CODECRAFTERS_SUBMISSION_DIR=/tmp/abc  \
-	CODECRAFTERS_TEST_CASES_JSON="[{\"slug\":\"repl-master-cmd-prop\",\"tester_log_prefix\":\"replication-11\",\"title\":\"Replication Stage\"}]" \
+	cd /Users/ryang/Developer/byox/build-your-own-redis && \
```
ryan-gang, line 27: Ignore the makefile, will revert changes in another commit.

rohitpaulk:
@ryan-gang you have test failures

internal/test_cases/receive_replication_handshake_test_case.go
```diff
@@ -62,17 +63,56 @@ func (t ReceiveReplicationHandshakeTestCase) RunReplconfStep1(client *resp_conne
...
 
-	return commandTest.Run(client, logger)
+	err := commandTest.Run(client, logger)
+	if err != nil {
+		return err
+	}
+
+	receivedValue := commandTest.ReceivedValue
+
+	if receivedValue.Type != resp_value.ARRAY {
```
rohitpaulk, line 78: When will this condition trigger? Doesn't `OnlyCommandAssertion` already test this?

internal/test_cases/receive_replication_handshake_test_case.go
```diff
@@ -62,17 +63,56 @@ func (t ReceiveReplicationHandshakeTestCase) RunReplconfStep1(client *resp_conne
...
+	elements := receivedValue.Array()
+
+	if len(elements) < 3 {
+		return fmt.Errorf("Expected array with at least 3 element, got %d elements", len(elements))
+	}
+
+	firstCapaArg := elements[1].String()
+
+	if elements[1].Type != resp_value.SIMPLE_STRING && elements[1].Type != resp_value.BULK_STRING {
+		return fmt.Errorf("Expected replconf argument to be a string, got %s", elements[1].Type)
```
rohitpaulk, line 91:
```suggestion
		return fmt.Errorf("Expected first replconf argument to be a string, got %s", elements[1].Type)
```

(Makes it more clear what you're testing)

internal/test_cases/receive_replication_handshake_test_case.go
```diff
@@ -62,17 +63,56 @@ func (t ReceiveReplicationHandshakeTestCase) RunReplconfStep1(client *resp_conne
...
+	}
+
+	if !strings.EqualFold(firstCapaArg, "capa") {
+		return fmt.Errorf("Expected replconf argument to be %q, got %q", strings.ToLower("capa"), strings.ToLower(firstCapaArg))
```
rohitpaulk, line 95:
```suggestion
		return fmt.Errorf("Expected replconf argument to be %q, got %q", "capa", strings.ToLower(firstCapaArg))
```

Not sure why the `strings.ToLower` is needed

internal/test_cases/receive_replication_handshake_test_case.go
```diff
@@ -62,17 +63,56 @@ func (t ReceiveReplicationHandshakeTestCase) RunReplconfStep1(client *resp_conne
...
+	}
+
+	if len(elements) == 5 {
+		secondCapaArg := elements[3].String()
```
rohitpaulk, line 99: Today, `resp_value.String()` doesn't cause troubles if called on a non-string resp value, but I can imagine cases where it does? Would place this after the type check below so that we're sure it's a string

internal/test_cases/send_command_test_case.go
```diff
@@ -0,0 +1,62 @@
...
+	Retries                   int
+}
+
+func (t SendCommandTestCase) Run(client *resp_client.RespConnection, logger *logger.Logger) error {
+	var value resp_value.Value
+	var err error
+
+	for attempt := 0; attempt <= t.Retries; attempt++ {
+		if attempt > 0 {
+			logger.Debugf("(Attempt %d/%d)", attempt, t.Retries)
```
rohitpaulk, line 27: Without seeing fixtures I can't tell what this'll look like, but based on a quick look this seems wrong - if there was 1 retry, looks like we'd log `Attempt 1/1`, and only on the "second" attempt? That could be misleading. Maybe this'd be clearer if we said `Retrying... (%d/%d attempts)`

internal/test_cases/send_command_test_case.go
```diff
@@ -0,0 +1,62 @@
...
+		if err = client.SendCommand(t.Command, t.Args...); err != nil {
+			return err
+		}
+
+		value, err = client.ReadValue()
+		if err != nil {
+			return err
+		}
+
+		if err := resp_assertions.NewNilAssertion().Run(value); err != nil {
```
rohitpaulk, line 39: Bad abstraction - when something is called "SendCommandTestCase", and it has `Retries`, I don't think it is clear that we're retrying based on whether the value is `nil`. You should accept something like `ShouldRetryFunc` from the caller and use that here. 

In this case, ShouldRetryFunc can be `resp_assertions.NewNilAssertion().run(value) == nil`. 

As an example, you might have another test where the value is currently "abcd", and you want to retry until the value is "efgh". 

In that case, `ShouldRetryFunc` can be `resp_assertions.NewStringAssertion(old_value).run(value) == nil`.

internal/test_repl_multiple_replicas.go
```diff
@@ -17,66 +18,63 @@ func testReplMultipleReplicas(stageHarness *test_case_harness.TestCaseHarness) e
...
+		}
 	}
 
-	for j := 0; j < 3; j++ {
+	// We then assert that across all the replicas we receive the SET commands in order
+	for j := 0; j < replicaCount; j++ {
 		replica := replicas[j]
 		logger.Infof("Testing Replica : %v", j+1)
+		for i := 1; i <= len(kvMap); i++ {
+			receiveCommandTestCase := &test_cases.ReceiveValueTestCase{
```
rohitpaulk, line 63: The variable here is named `receiveCommandTestCase`, but we're using a `ReceiveValueTestCase`, not a `ReceiveCommandTestCase`

internal/test_repl_multiple_replicas.go
```diff
@@ -17,66 +18,63 @@ func testReplMultipleReplicas(stageHarness *test_case_harness.TestCaseHarness) e
...
+			Assertion: resp_assertions.NewStringAssertion("OK"),
+		}
+		if err := setCommandTestCase.Run(client, logger); err != nil {
+			return err
+		}
 	}
 
-	for j := 0; j < 3; j++ {
+	// We then assert that across all the replicas we receive the SET commands in order
+	for j := 0; j < replicaCount; j++ {
```
rohitpaulk, line 59:
```suggestion
	for j, replica := range replicas
```

internal/test_repl_multiple_replicas.go
```diff
@@ -17,66 +18,63 @@ func testReplMultipleReplicas(stageHarness *test_case_harness.TestCaseHarness) e
...
 		replica := replicas[j]
 		logger.Infof("Testing Replica : %v", j+1)
```
rohitpaulk, line 61:
```suggestion
		logger.Infof("Testing Replica: %d/%d", j+1, replicaCount)
```

internal/test_repl_replica_cmd_processing.go
```diff
@@ -56,23 +66,26 @@ func testReplCmdProcessing(stageHarness *test_case_harness.TestCaseHarness) erro
 
 	for i := 1; i <= len(kvMap); i++ { // We need order of commands preserved
 		key, value := kvMap[i][0], kvMap[i][1]
-		err = master.Send([]string{"SET", key, value})
-		if err != nil {
+		// We are propagating commands to Replica as Master, don't expect any response back.
+		if err := master.SendCommand("SET", []string{key, value}...); err != nil {
```
rohitpaulk, line 70:
```suggestion
		if err := master.SendCommand("SET", key, value); err != nil {
```

internal/test_repl_replica_cmd_processing.go
```diff
@@ -56,23 +66,26 @@ func testReplCmdProcessing(stageHarness *test_case_harness.TestCaseHarness) erro
...
 			return err
 		}
-		// Master is propagating commands to Replica, don't expect any response back.
 	}
 
 	for i := 1; i <= len(kvMap); i++ {
 		key, value := kvMap[i][0], kvMap[i][1]
 		logger.Infof("Getting key %s", key)
```
rohitpaulk, line 77: I think we already have a log inside `SendCommandTestCase`, might not need this (can't tell since we don't have fixtures here)

internal/util.go
```diff
@@ -122,3 +128,23 @@ func IsSelectCommand(value resp_value.Value) bool {
 		value.Array()[0].Type == resp_value.BULK_STRING &&
 		strings.ToLower(value.Array()[0].String()) == "select"
 }
+func SpawnReplicas(replicaCount int, stageHarness *test_case_harness.TestCaseHarness, logger *logger.Logger, addr string) ([]*resp_connection.RespConnection, error) {
```
rohitpaulk, line 131: @ryan-gang let's add `gofmt` to the test workflow - there should be an empty line above this function if `gofmt` was run (correct me if I'm wrong)

rohitpaulk (approved):

rohitpaulk:
@ryan-gang mind tagging me once the fixtures are in place? would like to take a look at the final logs

internal/test_repl_replica_cmd_processing.go
ryan-gang:
```
[stage-113] Getting key foo
[stage-113] replica: $ redis-cli GET foo
[stage-113] replica: Sent bytes: "*2\r\n$3\r\nGET\r\n$3\r\nfoo\r\n"
[stage-113] replica: Received RESP value: "123"
[stage-113] Received "123"
```

internal/util.go
ryan-gang: @rohitpaulk `gofmt` didn't catch this. (VSCode and Goland are both running `gofmt` on file save).

I tried running `gofmt -s -l -w .` manually, it still doesn't change anything. 
Upon further investigation I found this: https://github.com/golang/go/issues/22337

ryan-gang:
@rohitpaulk fixtures updated. 
Tests passing.

PR #110, opened 2024-03-15
CC-1059: 5. Refactor handshake stages

ryan-gang:


internal/instrumented_resp_connection/instrumented_resp_connection.go
```diff
@@ -18,6 +18,9 @@ func defaultCallbacks(stageHarness *test_case_harness.TestCaseHarness, logPrefix
 				stageHarness.Logger.Infof("%s$ redis-cli %s", logPrefix, command)
 			}
 		},
+		BeforeSendValue: func(value resp_value.Value) {
+			stageHarness.Logger.Infof("%s$ %s", logPrefix, value.FormattedString())
```
rohitpaulk, line 22:
```suggestion
			stageHarness.Logger.Infof("%sSent: %s", logPrefix, value.FormattedString())
```

internal/resp/encoder/encode.go
```diff
@@ -41,6 +43,10 @@ func encodeBulkString(v resp_value.Value) []byte {
 	return []byte(fmt.Sprintf("$%d\r\n%s\r\n", len(v.Bytes()), v.Bytes()))
 }
 
+func encodeRDBAsBulkString(v resp_value.Value) []byte {
```
rohitpaulk, line 46:
```suggestion
func encodeFullResyncRDBFile(v resp_value.Value) []byte {
```

internal/resp/encoder/encode.go
```diff
@@ -15,6 +15,8 @@ func Encode(v resp_value.Value) []byte {
 		return encodeSimpleString(v)
 	case resp_value.BULK_STRING:
 		return encodeBulkString(v)
+	case resp_value.RDB_BULK_STRING:
```
rohitpaulk, line 18: Remove this, don't break the "RESPValue" abstraction

internal/resp/utils.go
```diff
@@ -0,0 +1,25 @@
...
+	const charset = "abcdefghijklmnopqrstuvwxyz0123456789"
+	result := make([]byte, length)
+	for i := 0; i < length; i++ {
+		charIndex := testerutils_random.RandomInt(0, len(charset))
+		result[i] = charset[charIndex]
+	}
+	return string(result)
+}
+
+func IsSelectCommand(value resp_value.Value) bool {
```
rohitpaulk, line 20: I'd move both of these out of `resp` package, keep them inside `internal`

internal/resp/utils.go
```diff
@@ -0,0 +1,25 @@
+package resp_utils
+
+import (
+	"strings"
+
+	resp_value "github.com/codecrafters-io/redis-tester/internal/resp/value"
+	testerutils_random "github.com/codecrafters-io/tester-utils/random"
+)
+
+func RandomAlphanumericString(length int) string {
```
rohitpaulk, line 10: Probably okay to use a fixed string, _unless_ we need to test/guarantee that the user's implementation can handle random strings

internal/resp/value/value.go
```diff
@@ -6,12 +6,13 @@ import (
...
-	SIMPLE_STRING string = "SIMPLE_STRING"
-	INTEGER       string = "INTEGER"
-	BULK_STRING   string = "BULK_STRING"
-	ARRAY         string = "ARRAY"
-	ERROR         string = "ERROR"
-	NIL           string = "NIL"
+	SIMPLE_STRING   string = "SIMPLE_STRING"
+	INTEGER         string = "INTEGER"
+	BULK_STRING     string = "BULK_STRING"
+	RDB_BULK_STRING string = "RDB_BULK_STRING"
```
rohitpaulk, line 12: Let's remove `RDB_BULK_STRING`, it breaks the idea of this being a RESP package - ideally it only contains types present in the Redis docs

internal/resp/value/value.go
```diff
@@ -37,6 +38,13 @@ func NewBulkStringValue(s string) Value {
 	}
 }
 
+func NewRDBAsBulkStringValue(s string) Value {
```
rohitpaulk, line 41: Can remove this too

internal/test_cases/receive_replication_handshake_test_case.go
```diff
@@ -0,0 +1,101 @@
...
+	return commandTest.Run(client, logger)
+}
+
+func (t ReceiveReplicationHandshakeTestCase) RunSendRDBStep(client *resp_connection.RespConnection, logger *logger.Logger) error {
+	logger.Debugln("Sending RDB file...")
+
+	hexStr := "524544495330303131fa0972656469732d76657205372e322e30fa0a72656469732d62697473c040fa056374696d65c26d08bc65fa08757365642d6d656dc2b0c41000fa08616f662d62617365c000fff06e3bfec0ff5aa2"
+	bytes, err := hex.DecodeString(hexStr)
+	if err != nil {
+		fmt.Printf("Encountered %s while deconding hex string", err.Error())
```
rohitpaulk, line 90: Just `panic` here - if there's `hex.MustDecodeString`, then use that

internal/test_cases/receive_replication_handshake_test_case.go
```diff
@@ -0,0 +1,101 @@
...
+		error := resp_value.NewErrorValue(err.Error())
+		client.SendBytes(error.Bytes())
+		return err
+	}
+
+	encodedValue := resp_encoder.Encode(resp_value.NewRDBAsBulkStringValue(string(bytes)))
```
rohitpaulk, line 96:
```suggestion
	encodedValue := resp_encoder.EncodeFullResyncRDBFile(bytes)
```

rohitpaulk (approved):
Blah

PR #112, opened 2024-03-15
CC-1059: 3. Update `resp_client` to `resp_connection`.

ryan-gang:
Instead of having different client and server packages, use a single shared connection package.


internal/resp/connection/connection.go
```diff
@@ -40,42 +40,50 @@ type RespClient struct {
...
-	return &RespClient{
+	return &RespConnection{
 		Conn:         conn,
 		UnreadBuffer: bytes.Buffer{},
 		Callbacks:    callbacks,
 	}, nil
 }
 
-func (c *RespClient) Close() error {
+func NewRespConnectionFromConnWithCallbacks(conn net.Conn, callbacks RespConnectionCallbacks) (*RespConnection, error) {
```
rohitpaulk, line 74: We can remvoe `WithCallbacks` from all these names, and also remove the `NewRespConnection` method (the one that doesn't take callbacks). 

It doesn't look like we use it at the moment, and if we have a need in the future - another way to do this would be to offer a `NewEmptyCallbacks()` function that returns a callbacks object with all functions filled that do nothing.

rohitpaulk (approved):

PR #114, opened 2024-03-22
feat: Update assertion interface & Introduce composite assertion.

ryan-gang:


internal/test_cases/send_command_test_case.go
```diff
@@ -45,10 +47,16 @@ func (t SendCommandTestCase) Run(client *resp_client.RespConnection, logger *log
 		}
 	}
 
-	if err = t.Assertion.Run(value); err != nil {
-		return err
+	result := t.Assertion.Run(value)
+	if result.IsFailure() {
+		if result.SuccessMessages != nil {
+			logger.Successf(strings.Join(result.SuccessMessages, "\n"))
+		}
```
ryan-gang, line 54: Even if we have a failure, we are printing out any success logs if present.
I have only added this in a single file. 
Looking for your feedback on this approach.

rohitpaulk:
@ryan-gang failing tests here too

rohitpaulk:
@ryan-gang I'll look at this once you have fixtures in place

ryan-gang:
Updated fixtures @rohitpaulk.

rohitpaulk:
Hmm, this feels a bit overkill - almost like we're exploding the interface just to satisfy one specific use-case. Lemme think about this a bit

rohitpaulk:
Closing in favour of https://github.com/codecrafters-io/redis-tester/pull/116

PR #116, opened 2024-03-25
Fix order-dependent test for replid stage

rohitpaulk:
The commit message could be: "Add new files and update existing files for Redis tester"

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/codecrafters-io/redis-tester
 go 1.21
 
 require (
-	github.com/codecrafters-io/tester-utils v0.2.16
+	github.com/codecrafters-io/tester-utils v0.2.17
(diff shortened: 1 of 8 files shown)


rohitpaulk:
There's something wrong with verifying fixtures - I think we have an overeager pattern somewhere that causes comparisons to go awry.

rohitpaulk:
Yep, regex fixed above. 

@ryan-gang let's go with something like this for the double-assertion case for now. We'll wait for 2-3 more use-cases to emerge, and then decide on a proper interface once we have those. I'm finding it a bit hard to come up with something clean/elegant at the moment, and that usually means that the specifications/use-cases are murky.

I know the `Received "<long_string>"` log isn't pretty, I'll think about fixes to that in a separate PR.

ryan-gang (approved):
Looks good.

PR #122, opened 2024-03-29
feat: improve bulk string parsing error logs when \r\n is missing at the end.

ryan-gang:
Fixes: https://github.com/codecrafters-io/build-your-own-redis/issues/157


rohitpaulk (changes requested):
Added a question

rohitpaulk:
Also, can you confirm what exactly this has to do with RDB? 

> Update RDB parsing error logs, to include how many bytes were read, and where the error occurred exactly.

The changes here are related to bulk strings, no?

rohitpaulk (approved):

PR #123, opened 2024-04-01
fix: Remove unnecessary escapes in response feedbacks from stream-6/7/8

andy1li:
**Problem:**

The final feedbacks from stream-6/7/8 contain unnecessary escapes instead of new lines.

**Solution:**

Copy code from stream-9, namely replacing `%q` with `%v`.

**Why:**

> %q: a single-quoted character literal safely escaped with Go syntax.
> %v: the value in a default format
> https://pkg.go.dev/fmt

**Forum:**
https://forum.codecrafters.io/t/stream-6-7-8-output-is-wacky-compared-to-stream-9/24


rohitpaulk:
Thanks for PR @andy1li!

(No idea why GitHub actions didn't trigger here, hmm... need to look into this)

rohitpaulk:
cc: @ryan-gang let's include this in the next release when you fix the publish job permissions please

ryan-gang:
Thanks for the PR @andy1li!

PR #128, opened 2024-04-05
Redis expiry test updated

NexFlare:
This pull request fixes following issue

Subsequent updates to a key's expiry time did not effectively overwrite the old expiration time with the new, more recent value. This behavior led to inconsistencies in key expiration, not aligning with user expectations or commands issued.

If we run
```
set key value px 100
```
and then immediately run
```
set key value px 200
```
the key should expire after 200ms

### DIFF
--- a/internal/test_expiry.go
+++ b/internal/test_expiry.go
@@ -55,6 +55,30 @@ func testExpiry(stageHarness *test_case_harness.TestCaseHarness) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


rohitpaulk:
Thanks for submitting this! Notes on why I'm closing this here: https://forum.codecrafters.io/t/redis-expiry-challenge-new-testcase-suggestion/44/3?u=rohitpaulk

PR #142, opened 2024-05-10
CC-1195: update --replicaof flag to pass single variable

ryan-gang:
- update tester
- update printed logs
- update stage descriptions

### DIFF
--- a/internal/redis_executable/redis_executable.go
+++ b/internal/redis_executable/redis_executable.go
@@ -1,8 +1,6 @@
 package redis_executable
 
 import (
-	"strings"
-
 	executable "github.com/codecrafters-io/tester-utils/executable"
 	logger "github.com/codecrafters-io/tester-utils/logger"
 	"github.com/codecrafters-io/tester-utils/test_case_harness"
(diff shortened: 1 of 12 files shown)


internal/redis_executable/redis_executable.go
```diff
@@ -30,7 +28,16 @@ func (b *RedisExecutable) Run(args ...string) error {
 	if b.args == nil || len(b.args) == 0 {
 		b.logger.Infof("$ ./spawn_redis_server.sh")
 	} else {
-		b.logger.Infof("$ ./spawn_redis_server.sh %s", strings.Join(b.args, " "))
+		var log string
+		log += "$ ./spawn_redis_server.sh"
+		for _, arg := range b.args {
+			if arg[0] == '-' {
```
rohitpaulk, line 34: Seems like a bit of a weird condition - can we just check if the arg has spaces in it instead? That way we don't unnecessarily quote other properties like "port" too

rohitpaulk (changes requested):

rohitpaulk (approved):

PR #143, opened 2024-05-10
CC-1182: add logs when we expect or wait for the user to do something

ryan-gang:
If the user fails to perform it, they wouldn't know what we were expecting. This way our expectations are clear.

### DIFF
--- a/internal/test_cases/receive_replication_handshake_test_case.go
+++ b/internal/test_cases/receive_replication_handshake_test_case.go
@@ -44,6 +44,9 @@ func (t ReceiveReplicationHandshakeTestCase) RunAll(client *resp_connection.Resp
 }
 
 func (t ReceiveReplicationHandshakeTestCase) RunPingStep(client *resp_connection.RespConnection, logger *logger.Logger) error {
+	logger.Debugf("Waiting for replica to initiate Replication handshake.")
+	logger.Debugf("Waiting for replica to send %q command", "PING")
+
 	commandTest := ReceiveCommandTestCase{
 		Assertion: resp_assertions.NewCommandAssertion("PING"),
 		Response:  resp_value.NewSimpleStringValue("PONG"),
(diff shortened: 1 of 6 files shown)


internal/test_cases/receive_replication_handshake_test_case.go
```diff
@@ -44,6 +44,9 @@ func (t ReceiveReplicationHandshakeTestCase) RunAll(client *resp_connection.Resp
 }
 
 func (t ReceiveReplicationHandshakeTestCase) RunPingStep(client *resp_connection.RespConnection, logger *logger.Logger) error {
+	logger.Debugf("master: Waiting for replica to initiate Replication handshake.")
+	logger.Debugf("master: Waiting for replica to send %q command", "PING")
```
rohitpaulk, line 48: @ryan-gang should these be debug or info?

internal/test_cases/receive_replication_handshake_test_case.go
```diff
@@ -53,6 +56,8 @@ func (t ReceiveReplicationHandshakeTestCase) RunPingStep(client *resp_connection
 }
 
 func (t ReceiveReplicationHandshakeTestCase) RunReplconfStep1(client *resp_connection.RespConnection, logger *logger.Logger) error {
+	logger.Debugf("master: Waiting for replica to send %q command", "REPLCONF listening-port")
```
rohitpaulk, line 59: Feels a bit weird to log out the partial command, can we log out the full version instead?

internal/test_cases/receive_replication_handshake_test_case.go
rohitpaulk: Maybe this can be shorter: `Waiting for replica to initiate handshake with "PING" command`

internal/test_repl_wait.go
```diff
@@ -131,6 +134,8 @@ func consumeReplicationStreamAndSendAcks(replicas []*resp_connection.RespConnect
 			}
 		}
 
+		logger.Infof("replica-%d: Expecting \"REPLCONF GETACK\" from Master", j+1)
```
rohitpaulk, line 137:
```suggestion
		logger.Infof("replica-%d: Expecting \"REPLCONF GETACK *\" from Master", j+1)
```

rohitpaulk (approved):

internal/test_repl_wait.go
```diff
@@ -131,6 +134,8 @@ func consumeReplicationStreamAndSendAcks(replicas []*resp_connection.RespConnect
...
+
```
rohitpaulk, line 138: @ryan-gang can we also add logs around when a replica decides to respond with an ACK vs. when it doesn't? This is a common point of confusion: https://forum.codecrafters.io/t/replication-stage-18-no-acks-received-from-replicas/63/5

internal/test_cases/receive_replication_handshake_test_case.go
ryan-gang: Info seems better suited, more so because of the color.

internal/test_repl_wait.go
ryan-gang: image

PR #148, opened 2024-05-28
CC-1208: add RESP data type to received data log

ryan-gang:


internal/instrumented_resp_connection/instrumented_resp_connection.go
```diff
@@ -28,7 +28,9 @@ func defaultCallbacks(stageHarness *test_case_harness.TestCaseHarness, logPrefix
 			stageHarness.Logger.Debugf("%sReceived bytes: %q", logPrefix, string(bytes))
 		},
 		AfterReadValue: func(value resp_value.Value) {
-			stageHarness.Logger.Debugf("%sReceived RESP value: %s", logPrefix, value.FormattedString())
+			valueTypeLowerCase := strings.ReplaceAll(strings.ToLower(value.Type), "_", " ")
```
rohitpaulk, line 31: For `NIL`, let's say either `null bulk string` or `nil (null bulk string)` - this has caused confusion for multiple users since they see the RESP3 null type.

internal/test_helpers/fixtures/expiry/pass
```diff
@@ -5,22 +5,22 @@ Debug = true
...
 [33m[stage-7] [0m[92mReceived "blueberry"[0m
 [33m[stage-7] [0m[36mSleeping for 101ms[0m
-[33m[stage-7] [0m[94mFetching key "apple" at 02:05:03.316 (should be expired)[0m
+[33m[stage-7] [0m[94mFetching key "apple" at 00:14:55.083 (should be expired)[0m
 [33m[stage-7] [0m[94m$ redis-cli GET apple[0m
 [33m[stage-7] [0m[36mSent bytes: "*2\r\n$3\r\nGET\r\n$5\r\napple\r\n"[0m
 [33m[stage-7] [0m[36mReceived bytes: "$-1\r\n"[0m
-[33m[stage-7] [0m[36mReceived RESP value: NIL[0m
-[33m[stage-7] [0m[92mReceived NIL[0m
+[33m[stage-7] [0m[36mReceived RESP nil: $-1\r\n[0m
```
rohitpaulk, line 22: Let's add quotes around the value? Looks a bit weird otherwise

rohitpaulk (approved):

PR #152, opened 2024-06-11
CC-1292: Update log prefix dynamically when reusing connection

ryan-gang:
This pull request updates the `BeforeSendCommand` callback in the `RespConnectionCallbacks` struct to include a new parameter `reusedConnection` which indicates whether the connection is being reused or not. The `BeforeSendCommand` callback is used to log the commands being sent to the server. With this change, the command prefix is dynamically set based on whether the connection is being reused or not. If the connection is reused, the command prefix is set to `">"`, otherwise it is set to "`$ redis-cli`". This provides more context in the logs when commands are being sent.

### DIFF
--- a/internal/resp/connection/connection.go
+++ b/internal/resp/connection/connection.go
@@ -14,7 +14,7 @@ import (
 type RespConnectionCallbacks struct {
 	// BeforeSendCommand is called when a command is sent to the server.
 	// This can be useful for info logs.
-	BeforeSendCommand func(command string, args ...string)
+	BeforeSendCommand func(reusedConnection bool, command string, args ...string)
 
 	// BeforeSendValue is called when a value is sent using SendValue.
 	// It is NOT called when using SendCommand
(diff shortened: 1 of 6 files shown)


ryan-gang (commented):
Fixes: https://github.com/codecrafters-io/build-your-own-redis/issues/129

internal/resp/connection/connection.go
```diff
@@ -14,7 +14,7 @@ import (
 type RespConnectionCallbacks struct {
 	// BeforeSendCommand is called when a command is sent to the server.
 	// This can be useful for info logs.
-	BeforeSendCommand func(command string, args ...string)
+	BeforeSendCommand func(reusedConnection bool, command string, args ...string)
```
rohitpaulk, line 17: Not a big deal since we only have this one use-case, but wanted to leave a note: 

A common pattern I've seen for cases like these is to pass in `RespConnection` itself to the callback - that way the callback has access to everything public on `RespConnection`, and it helps avoid parameters like `reusedConnection` which are use-case specific (over time we might have to add new parameters for multiple things that could just be computed off of `RespConnection`)

rohitpaulk (approved):

PR #153, opened 2024-06-12
CC-1299-txn

ryan-gang:


ryan-gang:
Lint is failing due to unused high level test functions. (The stage level methods).

internal/resp/value/value.go
```diff
@@ -70,7 +71,8 @@ func NewArrayValue(arr []Value) Value {
 
 func NewNilValue() Value {
 	return Value{
-		Type: NIL,
+		Type:  NIL,
+		bytes: []byte("$-1\r\n"),
```
rohitpaulk, line 75: Same as above, this shouldn't be "state" - it's derivable

internal/resp/value/value.go
```diff
@@ -41,6 +41,7 @@ func NewIntegerValue(i int) Value {
 	return Value{
 		Type:    INTEGER,
 		integer: i,
+		bytes:   []byte(fmt.Sprintf("%d", i)),
```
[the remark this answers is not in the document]
rohitpaulk: I see this as similar to `FormattedString()`. That's used for conveying what a value is to a user, and we just need a different representation specifically for comparisons. 

In fact, I wonder if we could just re-use `FormattedString()` for display purposes, and maybe add `Value.Equal(otherValue)` for comparison instead? 

i.e. we'd do something like: 

```
if !actualValue.Equal(expectedValue) {
  return fmt.Sprintf("expected %q, got %q", actualValue.formattedString(), expectedValue.formattedString())
}
```

internal/resp_assertions/ordered_array_assertion.go
```diff
@@ -0,0 +1,50 @@
...
+
+		if expectedValue.Bytes() == nil {
+			// This should never happen, but just in case
+			// This is the case for ArrayValues
+			return fmt.Errorf("CodeCrafters internal error. expectedValue Bytes of type: %s is nil", expectedValue.Type)
+		}
+
+		// ToDo: Equal or EqualFold ?
+		if !bytes.Equal(actualElement.Bytes(), expectedValue.Bytes()) {
+			return fmt.Errorf("Expected element #%d to be %s, got %s", i+1, expectedValue.FormattedString(), actualElement.FormattedString())
```
rohitpaulk, line 45: Okay, looks like we're already re-using `FormattedString()` here then (I left a comment re: this above). Would suggest implementing `Value.Equal(otherValue)` then - that will help clean this up, and will avoid the extra state stored in `bytes`

internal/test_cases/multi_command_test_case.go
```diff
@@ -0,0 +1,35 @@
...
+
+// MultiCommandTestCase is a concise & easier way to define & run multiple SendCommandTestCase
+type MultiCommandTestCase struct {
+	Commands   [][]string
+	Assertions []resp_assertions.RESPAssertion
+}
+
+func (t *MultiCommandTestCase) RunAll(client *resp_client.RespConnection, logger *logger.Logger) error {
+	if len(t.Assertions) != len(t.Commands) {
+		return fmt.Errorf("Number of commands and assertions should be equal")
```
rohitpaulk, line 19: @ryan-gang this should be an internal error

internal/test_cases/transaction_test_case.go
```diff
@@ -0,0 +1,79 @@
...
+		Args:      []string{},
+		Assertion: resp_assertions.NewStringAssertion("OK"),
+	}
+
+	return commandTest.Run(client, logger)
+}
+
+func (t TransactionTestCase) RunQueueAll(client *resp_client.RespConnection, logger *logger.Logger) error {
+	for i, v := range t.CommandQueue {
+		logger.Debugf("Sent #%d command", i)
```
rohitpaulk, line 57: This would read weirdly - it'd say "Sent 3 command". Can we do something like `Transaction: Sending command 1/3`?

internal/test_cases/transaction_test_case.go
```diff
@@ -0,0 +1,79 @@
...
+// can run each step individually.
+type TransactionTestCase struct {
+	// All the CommandQueue will be sent in order to client
+	// And a string "QUEUED" will be expected
+	CommandQueue [][]string
+
+	// After queueing all the commands,
+	// if ResultArray is not empty, "EXEC" is sent
+	// And the response is compared with this ResultArray
+	ResultArray []resp_value.Value
```
rohitpaulk, line 24: To be consistent with other test cases, I'd name this something like `ExpectedResponses`, or `ExpectedResponseArray`. 

- Include `Expected` because other test cases use this terminology
- Use `Response` because that's the "official" terminology for what a redis command returns. Either `Response` or `Value` works, `Result` seems like a new term we haven't used anywhere else

internal/resp_assertions/ordered_array_assertion.go
```diff
@@ -0,0 +1,50 @@
...
+func (a OrderedArrayAssertion) Run(value resp_value.Value) error {
+	if value.Type != resp_value.ARRAY {
+		return fmt.Errorf("Expected an array, got %s", value.Type)
+	}
+
+	if len(value.Array()) != len(a.ExpectedValue) {
+		return fmt.Errorf("Expected %d elements in array, got %d (%s)", len(a.ExpectedValue), len(value.Array()), value.FormattedString())
+	}
+
+	for i, expectedValue := range a.ExpectedValue {
```
rohitpaulk, line 30: @ryan-gang I think this should take assertions instead of values. Assertions already have a lot of friendliness built in to them, and we're losing all of that by not using them. We're essentially re-inventing a sub-par version of assertions here by trying to compare `Bytes`. 

For example, when you use a string assertion, by default it says: 

```
Expected simple string or bulk string, got %s
```

We could just use that, and wrap it to be something like `Element #4: <err>`?

rohitpaulk: Similarly, for `nil`: 

Screenshot 2024-06-13 at 12 49 44

With the current impelmentation, the error message would be `Expected element #2 to be a NIL, got XYZ`. Instead of the friendlier version that says "null bulk string" and includes `$-1\r\n`.

rohitpaulk (changes requested):
Notes added, primarily - we should be re-using assertions so that we get all the benefits they offer

internal/resp_assertions/ordered_array_assertion.go
ryan-gang: So, now we have no more requirement of the `Value.equal` / `Value.BytesForComparison` methods, right ?

ryan-gang:
@rohitpaulk, I have merged the stage implementations to this branch. 
To see how comments left on this PR specifically, have been fixed, please take a look at this diff: https://github.com/codecrafters-io/redis-tester/compare/6690867..a3b7777

internal/resp_assertions/error_assertion.go
```diff
@@ -0,0 +1,27 @@
...
+	return ErrorAssertion{ExpectedValue: expectedValue}
+}
+
+func (a ErrorAssertion) Run(value resp_value.Value) error {
+	if value.Type != resp_value.ERROR {
+		return fmt.Errorf("Expected error, got %s", value.Type)
+	}
+
+	if value.Error() != a.ExpectedValue {
+		return fmt.Errorf("Expected %s, got %s", a.ExpectedValue, value.Error())
```
rohitpaulk, line 23:
```suggestion
		return fmt.Errorf("Expected %q, got %q", a.ExpectedValue, value.Error())
```

(Quote, similar to string assertion)

internal/test_cases/transaction_test_case.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if err := t.RunQueueAll(client, logger); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (t TransactionTestCase) RunMulti(client *resp_client.RespConnection, logger *logger.Logger) error {
```
rohitpaulk, line 55: Ignore if these are used elsewhere, but to keep the interface clear, we could make `runMulti`, `runQueueAll` and `runExec` private so that's `Run` and `RunWithoutExec` stand out as the exposed methods

rohitpaulk, line 55: Ignore if these are used elsewhere, but to keep the interface clear, we could make `runMulti`, `runQueueAll` and `runExec` private so that's `Run` and `RunWithoutExec` stand out as the exposed methods.

rohitpaulk: Okay looks like `RunMulti` at least is used - and I'd assume `RunExec` is used too along with `RunWithoutExec`. Maybe `runQueueAll` can be private then?

internal/test_txn_tx_failure.go
```diff
@@ -0,0 +1,75 @@
...
+		return err
+	}
+
+	transactionTestCase := test_cases.TransactionTestCase{
+		CommandQueue: [][]string{
+			{"INCR", key1},
+			{"INCR", key2},
+		},
+		ExpectedResponseArray: []resp_assertions.RESPAssertion{
+			resp_assertions.NewErrorAssertion("ERR value is not an integer or out of range"), resp_assertions.NewIntegerAssertion(randomIntegerValue + 1)},
```
rohitpaulk, line 56: fyi you can format this better by adding a `,` after the last value, which will force one element per line

internal/test_cases/transaction_test_case.go
rohitpaulk: Might not make a big difference to make just one private, so resolving

rohitpaulk (approved):

PR #157, opened 2024-06-28
CC-1181: Add section-specific log prefixes to all stages

ryan-gang:
This pull request adds section-specific log prefixes to all stages in order to dynamically set section-specific log prefixes. This improves the readability and organization of the logs.

### DIFF
--- a/internal/instrumented_resp_connection/instrumented_resp_connection.go
+++ b/internal/instrumented_resp_connection/instrumented_resp_connection.go
@@ -9,7 +9,7 @@ import (
 	"github.com/codecrafters-io/tester-utils/logger"
 )
 
-func defaultCallbacks(Logger *logger.Logger, logPrefix string) resp_connection.RespConnectionCallbacks {
+func defaultCallbacks(logger *logger.Logger, logPrefix string) resp_connection.RespConnectionCallbacks {
 	return resp_connection.RespConnectionCallbacks{
 		BeforeSendCommand: func(reusedConnection bool, command string, args ...string) {
 			var commandPrefix string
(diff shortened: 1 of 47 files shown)


internal/test_repl_info_replica.go
```diff
@@ -51,7 +52,10 @@ func testReplInfoReplica(stageHarness *test_case_harness.TestCaseHarness) error
 		}
 		receiveReplicationHandshakeTestCase := test_cases.ReceiveReplicationHandshakeTestCase{}
 
+		logger.UpdateSecondaryPrefix("handshake")
```
rohitpaulk, line 55: Hmm, if this runs in a goroutine, couldn't it end up overridding the same logger used outside the goroutine?

rohitpaulk (changes requested):
Added q

internal/test_repl_info_replica.go
rohitpaulk: Question re-phrased - how do we prevent race conditions here?

internal/test_repl_info_replica.go
```diff
@@ -44,24 +45,27 @@ func testReplInfoReplica(stageHarness *test_case_harness.TestCaseHarness) error
 		defer conn.Close()
 
 		quietLogger := loggerutils.GetQuietLogger("")
-		master, err := instrumented_resp_connection.NewFromConn(stageHarness, conn, "master")
+		master, err := instrumented_resp_connection.NewFromConn(quietLogger, conn, "master")
```
ryan-gang, line 48: We can now pass the `quietLogger` directly to silence all callbacks.

ryan-gang: No need to update log context.

internal/instrumented_resp_connection/instrumented_resp_connection.go
```diff
@@ -6,10 +6,10 @@ import (
 
 	resp_connection "github.com/codecrafters-io/redis-tester/internal/resp/connection"
 	resp_value "github.com/codecrafters-io/redis-tester/internal/resp/value"
-	"github.com/codecrafters-io/tester-utils/test_case_harness"
+	"github.com/codecrafters-io/tester-utils/logger"
 )
 
-func defaultCallbacks(stageHarness *test_case_harness.TestCaseHarness, logPrefix string) resp_connection.RespConnectionCallbacks {
+func defaultCallbacks(Logger *logger.Logger, logPrefix string) resp_connection.RespConnectionCallbacks {
```
rohitpaulk, line 12:
```suggestion
func defaultCallbacks(logger *logger.Logger, logPrefix string) resp_connection.RespConnectionCallbacks {
```

(havent' seen us naming variables with capitals elsewhere)

rohitpaulk: Oh, is this to avoid conflicts with the package? In that case, I'd just import the package as `tester_utils_logger` or something like that. With this naming scheme it isn't clear why the variable is named differently.

internal/test_repl_info_replica.go
```diff
@@ -44,24 +45,27 @@ func testReplInfoReplica(stageHarness *test_case_harness.TestCaseHarness) error
...
 
-	client, err := instrumented_resp_connection.NewFromAddr(stageHarness, "localhost:6380", "client")
+	client, err := instrumented_resp_connection.NewFromAddr(logger, "localhost:6380", "client")
 	if err != nil {
 		logFriendlyError(logger, err)
 		return err
 	}
 	defer client.Close()
 
+	logger.UpdateSecondaryPrefix("test")
```
rohitpaulk, line 67: What's the prior prefix for other logs? Other stages have `handshake`, but I don't see that here

internal/test_repl_master_cmd_prop.go
```diff
@@ -55,6 +60,7 @@ func testReplMasterCmdProp(stageHarness *test_case_harness.TestCaseHarness) erro
 			Command:   "SET",
 			Args:      []string{key, value},
 			Assertion: resp_assertions.NewStringAssertion("OK"),
+			// Assertion: resp_assertions.NewStringAssertion("OK"),
```
rohitpaulk, line 63: Stray comment

rohitpaulk (approved):

PR #161, opened 2024-08-08
CC-1357: refactor logger.Infof calls for xread block tests

ryan-gang:
This pull request refactors the logger.Infof calls in the xread block tests to use formatted strings. This improves code readability and maintainability.

### DIFF
--- a/internal/test_helpers/fixtures/streams/pass
+++ b/internal/test_helpers/fixtures/streams/pass
@@ -4,7 +4,7 @@ Debug = true
 [33m[stage-44] [0m[94m$ ./spawn_redis_server.sh[0m
 [33m[stage-44] [0m[94m$ redis-cli xadd "pear" "0-1" "temperature 70"[0m
 [33m[stage-44] [0m[92mReceived response: ""0-1""[0m
-[33m[stage-44] [0m[94m$ redis-cli xread block '\x00' streams "pear 0-1"[0m
+[33m[stage-44] [0m[94m$ redis-cli xread block 0 streams "pear 0-1"[0m
 [33m[stage-44] [0m[94m$ redis-cli xadd "pear" "0-2" "temperature 70"[0m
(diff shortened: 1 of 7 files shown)


internal/test_helpers/fixtures/streams/pass
```diff
@@ -4,7 +4,7 @@ Debug = true
 [33m[stage-44] [0m[94m$ ./spawn_redis_server.sh[0m
 [33m[stage-44] [0m[94m$ redis-cli xadd "pear" "0-1" "temperature 70"[0m
 [33m[stage-44] [0m[92mReceived response: ""0-1""[0m
-[33m[stage-44] [0m[94m$ redis-cli xread block '\x00' streams "pear 0-1"[0m
+[33m[stage-44] [0m[94m$ redis-cli xread block 0 streams "pear 0-1"[0m
```
rohitpaulk, line 7: @ryan-gang can you check and see if "pear 0-1" is supposed to be quoted? I don't think it is: 

Screenshot 2024-08-08 at 15 12 20

rohitpaulk (changes requested):
Added note re: quoting

internal/test_helpers/fixtures/streams/pass
ryan-gang: Yep, shouldn't be quoted as per docs. 
But, as there is spacing inside, we kept it like this. Should I remove this then ?
This works with redis either way.

rohitpaulk: Yep let's remove quoting - seems like redis might have some kind of special handling for this if it works with quotes? The arguments seem to be separate arguments (one for key, one for id)

ryan-gang:
We use all unquoted params for `XREAD`, as per https://redis.io/docs/latest/commands/xread/.

ryan-gang:
Merging this in @rohitpaulk.

PR #164, opened 2024-09-03
Bump github.com/hdt3213/rdb to v1.0.18 for RDB version 11

andy1li:
In tandem with


rohitpaulk:
@andy1li let's add a test for RDBFileCreator that ensures the version number is what we expect - else we might end up mistakenly updating this again.

PR #165, opened 2024-09-03
Add version check test for RDB writer

ryan-gang:
This pull request adds a version check for the RDB writer. It includes a new test case that verifies the version of the RDB file created by the writer. The test ensures that the magic string is "REDIS" and the version is "0011". This check helps to ensure the correctness of the RDB writer implementation.


andy1li:
Looks great to me!

PR #168, opened 2024-10-17
refactor: update RespConnection struct to track total sent bytes

ryan-gang:
The SendCommand method has been updated to check the TotalSentBytes field instead of the SentBytes field to determine if a command is being sent for the first time or not.


andy1li:
Thanks, @leflings, for highlighting the issue!

PR #175, opened 2025-04-29
CC-1731: Fix Redis #xu1

andy1li:
- Add a **XReadResponseAssertion** to utilize `instrumented_resp_connection.NewFromAddr`
- Extend decodeArray to support **null array**($-1\r\n)

New fixture:
image

Error case:
image

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of nil arrays in responses, ensuring correct behavior when a length of -1 is encountered.

- **Tests**
	- Refactored stream reading tests for better structure and reliability.
	- Updated test fixtures with new key names, values, and enhanced protocol-level logging for improved clarity and accuracy.
	- Synchronized test data and expected outputs with updated Redis versions and replication details.

- **New Features**
	- Added new test assertions for validating XREAD command responses, enabling more precise and detailed test validations.


internal/test_streams_xread_block_max_id.go
```diff
@@ -22,93 +20,88 @@ func testStreamsXreadBlockMaxID(stageHarness *test_case_harness.TestCaseHarness)
...
-		}).Result()
+		expectedValue := [][]interface{}{
+			{
+				randomKey,
+				[]interface{}{
+					[]interface{}{"0-2", []interface{}{"temperature", strconv.Itoa(randomInt)}},
+				},
+			},
+		}
+		assertion := resp_assertions.NewStreamAssertion(expectedValue)
```
rohitpaulk, line 57: @andy1li it's generally considered bad to have to deal with `interface{}` for calls where the structure is actually known. It makes sense when dealing with freeform JSON from the user's code (which you might not know the structure for beforehand). In cases where the structure is known, this is just opens up the possibilities for multiple errors in the way the value is used (for example someone might forget the outer grouping, or add an additional nesting level).

rohitpaulk (changes requested):
Added notes around avoiding `interface{}{}` usage for known structures

internal/resp_assertions/xread_response_assertion.go
```diff
@@ -0,0 +1,102 @@
...
+
+func normalizeActual(v resp_value.Value) interface{} {
+	switch v.Type {
+	case resp_value.BULK_STRING:
+		return v.String()
+	case resp_value.ARRAY:
+		arr := v.Array()
+		result := make([]interface{}, len(arr))
+		for i, elem := range arr {
+			// Special handling for stream entries (second element of the outer array)
```
rohitpaulk, line 78: Looks good overall but I don't understand why we need this - let's go over on today's call!

rohitpaulk (approved):

PR #177, opened 2025-05-12
WIP

andy1li:


internal/test_cases/send_command_with_sync_test_case.go
```diff
@@ -0,0 +1,84 @@
...
+	Assertion                 resp_assertions.RESPAssertion
+	ShouldSkipUnreadDataCheck bool
+	Retries                   int
+	ShouldRetryFunc           func(resp_value.Value) bool
+
+	// ReceivedResponse is set after the test case is run
+	ReceivedResponse resp_value.Value
+
+	WaitChan chan bool
+	DoneChan chan bool
```
rohitpaulk, line 26: We shouldn't have to pass in `DoneChan`, a function already notifies you when it's done - the caller can manage the channel on their end if required.

rohitpaulk (changes requested):
comments added

rohitpaulk (approved):

.github/workflows/test.yml
```diff
@@ -45,3 +45,26 @@ jobs:
         with:
           version: "2023.1"
           install-go: false
+
+  test-flakiness:
```
rohitpaulk, line 49: @andy1li let's remove this for now - not worth keeping in CI unless this is a frequent occurrence, let's keep CI fast. We can use the `Makefile` test when needed for now

PR #180, opened 2025-06-19
fix: correct mutex handling in SendCommandTestCase Run method

andy1li:
Context:

https://forum.codecrafters.io/t/sending-output-but-not-being-recognised-by-tester/12346


andy1li:
Thanks to @EshaanAgg for highlighting the issue!

PR #183, opened 2025-06-29
Upgrade hdt3213/rdb to eliminate trailing LF in RDB

andy1li:


andy1li:
Thanks to @kelduben-cc and @edgerunner for highlighting the issue!

PR #184, opened 2025-07-01
Redis Lists Extension

UdeshyaDhungana:
- Added test cases for redis lists extension
- Update course definition


internal/test_cases/util.go
```diff
@@ -0,0 +1,11 @@
+package test_cases
+
+func Filter[T any](s []T, predicate func(T) bool) []T {
```
rohitpaulk, line 3: Let's use https://github.com/samber/lo/blob/master/README.md for this

internal/test_list_llen.go
```diff
@@ -0,0 +1,44 @@
...
+
+	randomKey := testerutils_random.RandomWord()
+	listSize := testerutils_random.RandomInt(4, 8)
+	randomList := testerutils_random.RandomWords(listSize)
+
+	multiCommandTestCase := test_cases.MultiCommandTestCase{
+		Commands: [][]string{
+			append([]string{"RPUSH", randomKey}, randomList...),
+			{"LLEN", randomKey},
+			{"LLEN", "non_existent_list_key"},
```
rohitpaulk, line 34:
```suggestion
			{"LLEN", "missing_key_#{random_integer}"},
```

(Let's add a bit of randomness here + use a simpler word "missing" instead of "non_existent")

internal/test_list_lpop2.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		logFriendlyError(logger, err)
+		return err
+	}
+	defer client.Close()
+
+	randomListKey := testerutils_random.RandomWord()
+	listSize := testerutils_random.RandomInt(5, 9)
+	elements := testerutils_random.RandomWords(listSize)
+	toRemoveCount := testerutils_random.RandomInt(2, 5)
```
rohitpaulk, line 31: Just confirming - there's no chance that listSize will be 5 and we'll remove 5 too, yes? Not sure if OrderedStringArrayAssertion will work well with empty arrays

internal/test_list_lpush.go
```diff
@@ -0,0 +1,45 @@
...
+	multiCommandTestCase := test_cases.MultiCommandTestCase{
+		Commands: [][]string{
+			{"LPUSH", randomListKey, randomValues[2]},
+			{"LPUSH", randomListKey, randomValues[1], randomValues[0]},
+			{"LRANGE", randomListKey, strconv.Itoa(0), strconv.Itoa(-1)},
+		},
+		Assertions: []resp_assertions.RESPAssertion{
+			resp_assertions.NewIntegerAssertion(1),
+			resp_assertions.NewIntegerAssertion(3),
+			resp_assertions.NewOrderedStringArrayAssertion([]string{randomValues[0], randomValues[1], randomValues[2]}),
```
rohitpaulk, line 40:
```suggestion
			resp_assertions.NewOrderedStringArrayAssertion(randomValues),
```

(Think this will work?)

internal/test_list_lrange_neg_index.go
```diff
@@ -0,0 +1,64 @@
...
+	client, err := instrumented_resp_connection.NewFromAddr(logger, "localhost:6379", "client")
+	if err != nil {
+		logFriendlyError(logger, err)
+		return err
+	}
+	defer client.Close()
+
+	randomListKey := testerutils_random.RandomWord()
+	listSize := testerutils_random.RandomInt(4, 5)
+	randomList := testerutils_random.RandomWords(listSize)
```
rohitpaulk, line 30: We name this differently in different stages - in some it's called "elements", in some it's called "list". I think elements is a more obvious name

```suggestion
	randomElements := testerutils_random.RandomWords(listSize)
```

internal/test_list_lrange_pos_index.go
```diff
@@ -0,0 +1,63 @@
...
+			{"LRANGE", randomListKey, "0", strconv.Itoa(listSize - 1)},
+
+			// start index > end index
+			{"LRANGE", randomListKey, "1", "0"},
+
+			// end index out of bounds
+			{"LRANGE", randomListKey, "0", strconv.Itoa(listSize * 2)},
+
+			// key doesn't exist
+			{"LRANGE", "non_existent_key", "0", "1"},
```
rohitpaulk, line 49:
```suggestion
			{"LRANGE", "missing_key_#{random_int}", "0", "1"},
```

(Consistent with comment above)

internal/test_list_rpush1.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	defer client.Close()
+
+	keyAndValue := testerutils_random.RandomWords(2)
+
+	testcase := test_cases.SendCommandTestCase{
+		Command:                   "RPUSH",
+		Args:                      keyAndValue,
+		Assertion:                 resp_assertions.NewIntegerAssertion(1),
+		ShouldSkipUnreadDataCheck: false,
```
rohitpaulk, line 32: `false` is the default, so we can ignore this line entirely

internal/test_list_blpop_no_timeout.go
```diff
@@ -0,0 +1,65 @@
...
+					Assertion: blPopAssertion,
+				},
+				ExpectResult: true,
+			},
+			{
+				Client: clients[1],
+				SendCommandTestCase: &test_cases.SendCommandTestCase{
+					Command:   "BLPOP",
+					Args:      []string{listKey, "0"},
+					Assertion: blPopAssertion,
```
rohitpaulk, line 48: It's kinda weird that we have to pass in an assertion but not use it if `ExpectResult` is false. Let's think of a better way around this?

rohitpaulk: I think we're indexing a bit too much on re-using `SendCommandTestCase`, which was intended for sending a command and receiving a response right away.

internal/test_cases/blocking_command_test_case.go
```diff
@@ -0,0 +1,120 @@
...
+	})
+
+	errorChan := make(chan error, len(idleClientsTestCase))
+	doneChan := make(chan bool, len(idleClientsTestCase))
+
+	for _, tc := range idleClientsTestCase {
+		go func(tc *ClientTestCase) {
+			defer func() { doneChan <- true }()
+
+			tc.PauseReadingResponse()
```
rohitpaulk, line 85: Not sure why this is needed, we aren't running `tc.Run` anywhere, so `tc.readMutex` isn't even being used no?

internal/test_list_blpop_no_timeout.go
rohitpaulk: @UdeshyaDhungana let's take another stab at coming up with a better structure for this -- happy to help with a starting point if you're stuck, but let's give this a go on your own first. I think the core mistake here is to try to rely on `SendCommandTestCase`, try removing that and things will become cleaner

rohitpaulk (changes requested):
Notes added!

internal/test_list_lpop2.go
UdeshyaDhungana: I tested it with 
```
listSize := 5
toRemoveCount := 5
```
It works with empty arrays as well.

internal/test_cases/blocking_command_test_case.go
UdeshyaDhungana: Sorry for this one. I should have used a single mutex for the entire test case instead.

internal/test_list_blpop_no_timeout.go
rohitpaulk: Primarily the latter - making the API cleaner. 

Based on a cursory look I thought the reason the API isn't friendly is that we were trying to re-use `SendCommandTestCase`, hence the suggestion to focus on removing that here. Happy to keep it in if we can still figure out a way to keep the outward API for `BlockingCommandTestCase` clean!

stage_descriptions/lists_8_ef1.md
```diff
@@ -0,0 +1,46 @@
+In this stage, you'll implement support for removing a single element from the left using the `LPOP` command.
```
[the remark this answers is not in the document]
UdeshyaDhungana: Got it!

Moved them from here to `build-your-own-redis`.

Ref: https://github.com/codecrafters-io/build-your-own-redis/pull/324

PR #187, opened 2025-07-03
Refactor: Avoid random in naming and re-use `ReceiveValueTestCase`

UdeshyaDhungana:
- Removed usage of `random` in variable names
- Re-used `ReceiveValueTestCase` inside `SendCommandTestCase`
- Updated the usage of `MkdirTemp` and updated fixtures (Reason below)

- Github Tests were failing randomly because of the following reason:

We are currently using the pretty library with options of width = 32, which means while printing, if the line is longer than 32 characters, it will split the json into separate line. I have experimented with different widths, and the logs looks readable at this width

In RDB test case ZG5, the `dir` parameter is a temporary directory, which is sth like `/var/folders/sn/8vpsm6lx6m53blpdh2yhlzw40000gn/T/`, which is pretty long compared to `/tmp/` in Linux. In case of MacOS, the length of temp directory is long, which automatically makes the array `["dir", "<temp_dir_name>"]` longer than pretty's width option, so it is splitted into different lines when printed.

However, in case of linux, whether or not that array will be longer than Pretty's width is random. `MkdirTemp` depends on this function:

```go
func nextRandom() string {
	return itoa.Uitoa(uint(uint32(runtime_rand())))
}
```

Due to this, the tests were randomly failing. For eg. you can see that the test passes for this PR #186 , but when it is merged, the tests fail. You can see this in the [repo homepage](https://github.com/codecrafters-io/redis-tester)

Screenshot 2025-07-03 at 17 57 04

So, I have changed the usage of `MkdirTemp` from

```
tmpDir, err := os.MkdirTemp("", "rdbfiles")
```

to

```
tmpDir, err := os.MkdirTemp("", "codecrafters-rdbfiles-")
```

This will cause the length of smallest possible length such array to be 39.

```python
>>> len('["dir", "/tmp/codecrafters-rdbfiles-0"]')
39
```
This will cause the array in that test case to be printed into separate lines, and will become consistent with the fixtures in macos.

### DIFF
--- a/internal/rdb_file_creator.go
+++ b/internal/rdb_file_creator.go
@@ -26,7 +26,7 @@ type RDBFileCreator struct {
 }
 
 func NewRDBFileCreator() (*RDBFileCreator, error) {
-	tmpDir, err := os.MkdirTemp("", "codecrafters-rdbfiles")
+	tmpDir, err := os.MkdirTemp("", "codecrafters-rdbfiles-")
 	if err != nil {
 		return &RDBFileCreator{}, err
 	}
--- a/internal/rdb_file_creator.go
+++ b/internal/rdb_file_creator.go
@@ -26,7 +26,7 @@ type RDBFileCreator struct {
 }
 
 func NewRDBFileCreator() (*RDBFileCreator, error) {
-	tmpDir, err := MkdirTemp("rdbfiles")
+	tmpDir, err := MkdirTemp("rdb")
 	if err != nil {
 		return &RDBFileCreator{}, err
 	}
--- a/internal/rdb_file_creator.go
+++ b/internal/rdb_file_creator.go
@@ -26,7 +26,7 @@ type RDBFileCreator struct {
 }
 
 func NewRDBFileCreator() (*RDBFileCreator, error) {
-	tmpDir, err := os.MkdirTemp("", "rdbfiles")
+	tmpDir, err := os.MkdirTemp("", "codecrafters-rdbfiles")
 	if err != nil {
 		return &RDBFileCreator{}, err
 	}
--- a/internal/rdb_file_creator.go
+++ b/internal/rdb_file_creator.go
@@ -26,19 +26,13 @@ type RDBFileCreator struct {
 }
 
 func NewRDBFileCreator() (*RDBFileCreator, error) {
-	tmpDir, err := os.MkdirTemp("", "codecrafters-rdbfiles-")
+	tmpDir, err := MkdirTemp("rdbfiles")
 	if err != nil {
 		return &RDBFileCreator{}, err
 	}
 
-	// On MacOS, the tmpDir is a symlink to a directory in /var/folders/...
-	realDir, err := filepath.EvalSymlinks(tmpDir)
-	if err != nil {
-		return &RDBFileCreator{}, fmt.Errorf("could not resolve symlink: %v", err)
-	}
-
 	return &RDBFileCreator{
-		Dir:      realDir,
+		Dir:      tmpDir,
 		Filename: fmt.Sprintf("%s.rdb", random.RandomWord()),
 	}, nil
 }
--- a/internal/test_cases/receive_value_test_case.go
+++ b/internal/test_cases/receive_value_test_case.go
@@ -22,13 +22,12 @@ func (t *ReceiveValueTestCase) Run(client *resp_client.RespConnection, logger *l
 	if err != nil {
 		return err
 	}
-	return t.runAssertionAndCheck(client, logger, value)
+	t.ActualValue = value
+	return t.runAssertionAndCheck(client, logger)
 }
 
-func (t *ReceiveValueTestCase) runAssertionAndCheck(client *resp_client.RespConnection, logger *logger.Logger, value resp_value.Value) error {
-	t.ActualValue = value
-
-	if err := t.Assertion.Run(value); err != nil {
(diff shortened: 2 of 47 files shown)


internal/rdb_file_creator.go
```diff
@@ -26,7 +26,7 @@ type RDBFileCreator struct {
 }
 
 func NewRDBFileCreator() (*RDBFileCreator, error) {
-	tmpDir, err := os.MkdirTemp("", "rdbfiles")
+	tmpDir, err := os.MkdirTemp("", "codecrafters-rdbfiles-")
```
rohitpaulk, line 29: @UdeshyaDhungana is there a specific reason to change this? If not let's not do it - shorter file paths are easier to read.

internal/test_rdb_config.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 func testRdbConfig(stageHarness *test_case_harness.TestCaseHarness) error {
-	tmpDir, err := os.MkdirTemp("", "rdbfiles")
+	tmpDir, err := os.MkdirTemp("", "codecrafters-rdbfiles-")
```
rohitpaulk, line 18: Same note as above

internal/test_cases/receive_value_test_case.go
```diff
@@ -22,10 +22,12 @@ func (t *ReceiveValueTestCase) Run(client *resp_client.RespConnection, logger *l
 	if err != nil {
 		return err
 	}
-
 	t.ActualValue = value
+	return t.AssertAndCheckUnread(client, logger)
+}
 
-	if err = t.Assertion.Run(value); err != nil {
+func (t *ReceiveValueTestCase) AssertAndCheckUnread(client *resp_client.RespConnection, logger *logger.Logger) error {
```
rohitpaulk, line 29:
```suggestion
func (t *ReceiveValueTestCase) Assert(client *resp_client.RespConnection, logger *logger.Logger) error {
```

(We might not actually be checking for unread here if `ShouldSkipUnreadDataCheck` is set)

internal/test_cases/send_command_test_case.go
```diff
@@ -66,21 +66,12 @@ func (t *SendCommandTestCase) Run(client *resp_client.RespConnection, logger *lo
...
-		}
+	receiveCommandTestCase := ReceiveValueTestCase{
+		Assertion:                 t.Assertion,
+		ShouldSkipUnreadDataCheck: t.ShouldSkipUnreadDataCheck,
+		ActualValue:               value,
 	}
-
-	logger.Successf("Received %s", value.FormattedString())
-	return nil
+	return receiveCommandTestCase.AssertAndCheckUnread(client, logger)
```
rohitpaulk, line 74: To make it a bit more clear that we're essentially re-using `ReceiveCommandTestCase` but splitting it into two to avoid asserting early, I'd do something like this: 

- Add `ReceiveCommandTestCase#RunWithoutAssert()`, use that and `#Assert()` internally in `Run`
- Use that instead of `client.ReadValue()` above, and then call `#Assert()` at the end here.

rohitpaulk (approved):

internal/rdb_file_creator.go
UdeshyaDhungana: @rohitpaulk  I have pointed out the issue with failing tests [here](https://github.com/codecrafters-io/redis-tester/pull/187#issue-3198958130).

Another way we could do this is use a separate value of width parameter while running tests. This would change the fixtures as well.

Please let me know which one would you prefer.

rohitpaulk: Ah, sorry for not checking the description! From a user's perspective, I prefer the single-line shorter directory version. That said I don't think we need to alter width altogether in tests, that seems like we're straying from what a real user would see and might not catch certain bugs. 

How about just creating a directory named `/tmp/rdbfiles-<randomInt>` and adding a comment to say we're doing this to keep fixtures consistent across macOS and Linux?

rohitpaulk:
Replied to the rdbfiles note!

internal/rdb_file_creator.go
UdeshyaDhungana: Sure!

UdeshyaDhungana:
- Revert tempdir to `rdbfiles`
- Use random number instead of `MkdirTemp`, the range is above 1,000,000 to produce consistent fixtures
- Split `ReceiveCommandTestCase` as requested

internal/util.go
```diff
@@ -155,3 +158,23 @@ func FormatKeyValuePairs(keys []string, values []string) string {
 
 	return fmt.Sprintf("{%s}", strings.Join(pairs, ", "))
 }
+
+// MkdirTemp keeps fixtures consistent across macOS and Linux
+func MkdirTemp(prefix string) (string, error) {
+	// ensure the length of tmpDir is long enough to be printed on multiple lines
```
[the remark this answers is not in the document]
rohitpaulk: Oh, okay. Just curious - what's the minimum length of randomint that would cause everything to fit on one line?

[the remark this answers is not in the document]
UdeshyaDhungana: If we account for MacOS as well,

```
>>> len('["dir", "/private/tmp/rdbfiles-0"]')
34
```

There's no minimum limit since the smallest value will be of length 34. (limit is 32 for single line).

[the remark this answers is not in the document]
rohitpaulk: Ah, okay - how about doing this? `["dir", "/private/tmp/rdb-1234"]`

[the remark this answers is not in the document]
UdeshyaDhungana: Hmm that could be done!

PR #189, opened 2025-07-09
Update: Logger Improvements, Test Cases Update

UdeshyaDhungana:
### Packages
- Update tester utils version

### Logging Improvements
- Connections are logged as secondary prefixes
- All connections are updated to have connection identifier
- Update fixtures
- Cases like `Syntax Error` now includes 'Received Bytes` line

### Test Cases Update
- Update blocking `XREAD` stages to use `BlockingClientGroupTestCase`
- Update `ReceiveCommandTestCase` to re-use `ReceiveValueTestCase`

### DIFF
--- a/internal/resp/connection/connection.go
+++ b/internal/resp/connection/connection.go
@@ -77,11 +77,12 @@ func NewRespConnectionFromAddr(addr string, connIdentifier string, callbacks Res
 	}, nil
 }
 
-func NewRespConnectionFromConn(conn net.Conn, callbacks RespConnectionCallbacks) (*RespConnection, error) {
+func NewRespConnectionFromConn(conn net.Conn, callbacks RespConnectionCallbacks, connID string) (*RespConnection, error) {
 	return &RespConnection{
 		Conn:         conn,
 		UnreadBuffer: bytes.Buffer{},
 		Callbacks:    callbacks,
+		identifier:   connID,
 	}, nil
 }
 
--- a/internal/resp/connection/connection.go
+++ b/internal/resp/connection/connection.go
@@ -152,8 +152,12 @@ func (c *RespConnection) ReadFullResyncRDBFile() ([]byte, error) {
 
 	value, readBytesCount, err := resp_decoder.DecodeFullResyncRDBFile(c.UnreadBuffer.Bytes())
 
-	if c.Callbacks.AfterBytesReceived != nil && readBytesCount > 0 {
-		c.Callbacks.AfterBytesReceived(c.UnreadBuffer.Bytes()[:readBytesCount])
+	if c.Callbacks.AfterBytesReceived != nil {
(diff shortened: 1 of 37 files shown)


UdeshyaDhungana:
@rohitpaulk I've made the changes. Ready for review!

internal/resp/decoder/errors.go
```diff
@@ -18,6 +18,11 @@ type InvalidInputError struct {
 	Message string
 }
 
+type InvalidRESPError struct {
```
rohitpaulk, line 21: @UdeshyaDhungana how is `InvalidRESPError` different from `InvalidInputError`? The idea behind the errors here was that there are two ways in which we can fail to parse RESP: either (a) the input is invalid (i.e. a protocol error), or (b) the input is incomplete (valid so far but incomplete without more bytes). InvalidInputError was supposed to represent (a). This seems like the same thing, yes?

internal/resp/connection/connection.go
```diff
@@ -77,11 +77,12 @@ func NewRespConnectionFromAddr(addr string, connIdentifier string, callbacks Res
 	}, nil
 }
 
-func NewRespConnectionFromConn(conn net.Conn, callbacks RespConnectionCallbacks) (*RespConnection, error) {
+func NewRespConnectionFromConn(conn net.Conn, callbacks RespConnectionCallbacks, connID string) (*RespConnection, error) {
 	return &RespConnection{
 		Conn:         conn,
 		UnreadBuffer: bytes.Buffer{},
 		Callbacks:    callbacks,
+		identifier:   connID,
```
rohitpaulk, line 85: Taking another look at this, the reason we split `instrumented_resp_connection` and added callbacks here was to limit scope / separate concerns: to keep this focused on the protocol and to keep `instrumented_resp_connection` focused on tracking things like client identifier, logging prefix etc. Our alternative was to inject logger directly here and have it handle all the logging, that gets messy pretty quickly given that this struct already does a bunch of work.  

What are your thoughts on creating an `InstrumentedRespConnection` type, and moving `Identifier` there?

rohitpaulk (changes requested):
Added one round of comments! One suggestion to split this PR: 

- One PR for upgrading tester-utils and just renaming usages
- Another PR to actually change the logging format wherever necessary

internal/resp/connection/connection.go
UdeshyaDhungana: Ahh, I see! Sure, will go with this approach.

internal/resp/decoder/errors.go
UdeshyaDhungana: My idea was to classify the errors into a third case where the input does not even begin with valid RESP marker (`+`, `-`, `:`, `$`, `*`). 
But yeah, this does seem redundant. I'll go over and change this.

PR #191, opened 2025-07-10
Change logging format

UdeshyaDhungana:
## Part 2/2 from #189 

- Currently using `logger-update` branch from tester utils
- Will need to revert to main branch if the change is merged

---

- Changed to c`lient.GetLogger()` format
- Removed `InvalidRespError` and re-used `InvalidInputError`
- `Received bytes...` will be shown for all cases

I think we should not store logger in `InstrumentedRespConnection`.  Many stages like `testReplMasterCmdProp` use `logger.UpdateSecondaryPrefix()` in between interleaved with test cases. It'll introduce extra cost of updating prefixes of connection's logger once the main logger is changed.

So, created a method with signature `InstrumentedRespConnection.GetLogger(*logger Logger) *logger.Logger` that returns a temporary logger (derived from main logger) with attached suffix.

https://github.com/codecrafters-io/tester-utils/pull/63 is a part of this.


rohitpaulk:
@UdeshyaDhungana feel free to split this so that it goes through review the quicker - the logging on `InvalidRespError` should be a simple change, and we should add a test for it so that we don't have regressions. That can be a PR in itself.

UdeshyaDhungana:
Update:
- `InvalidRespError` is not handled
- Logging changed
- Blocking XREAD stages now use BlockingClientGroupTestCase`

internal/instrumented_resp_connection/instrumented_resp_connection.go
```diff
@@ -20,48 +27,69 @@ func defaultCallbacks(logger *logger.Logger, logPrefix string) resp_connection.R
...
+// GetLogger returns a new logger with added secondary prefix: connection's identifier
+func (c *InstrumentedRespConnection) GetLogger() *logger.Logger {
+	return c.logger
+}
 
-	return resp_connection.NewRespConnectionFromConn(
-		conn, defaultCallbacks(logger, logPrefix),
-	)
+// UpdateLogger updates the connection's logger
+func (c *InstrumentedRespConnection) UpdateLogger(l *logger.Logger) {
```
rohitpaulk, line 90:
```suggestion
func (c *InstrumentedRespConnection) UpdateBaseLogger(l *logger.Logger) {
```

It could be confusing how we pass in a logger and a secondary prefix is automatically added to it. Looking at `GetLogger()` and `UpdateLogger(logger)`, I'd expect both to be symmetrical, i.e. after `UpdateLogger(logger)` is called with a logger, I'd get the same back in `GetLogger()`. Let's use the term `baseLogger` to make this distinction

rohitpaulk (approved):

PR #195, opened 2025-07-11
Blocking Client Group Test Case: Logging Improvements

UdeshyaDhungana:
Ref: https://forum.codecrafters.io/t/bug-in-tester-list-blpop/13315

- Introduced more logging to make it obvious what we're doing


internal/test_helpers/fixtures/lists/pass
```diff
@@ -14,9 +14,11 @@ Debug = true
 [33m[tester::#XJ7] [client-2] [0m[36mReceived bytes: ":1\r\n"[0m
 [33m[tester::#XJ7] [client-2] [0m[36mReceived RESP integer: 1[0m
 [33m[tester::#XJ7] [client-2] [0m[92mReceived 1[0m
+[33m[tester::#XJ7] [client-1] [0m[94mExpecting reply from server[0m
 [33m[tester::#XJ7] [client-1] [0m[36mReceived bytes: "*2\r\n$9\r\nraspberry\r\n$5\r\nmango\r\n"[0m
 [33m[tester::#XJ7] [client-1] [0m[36mReceived RESP array: ["raspberry", "mango"][0m
 [33m[tester::#XJ7] [client-1] [0m[92mReceived ["raspberry", "mango"][0m
+[33m[tester::#XJ7] [client-1] [0m[92mSuccess; received expected response[0m
```
rohitpaulk, line 21: The above log is already a success log, so this is pretty much redundant, would remove. @UdeshyaDhungana I'd think a bit harder about what's the exact case where our logs aren't conveying sufficient information, add a fixture for that case and then optimize from there. I see the linked forum post - let's get that into a fixture, that way we can see exactly what logs look like for that case, and what we need to tweak.

rohitpaulk (changes requested):
Added note

internal/test_cases/blocking_client_group_test_case.go
```diff
@@ -53,12 +53,16 @@ func (t *BlockingClientGroupTestCase) SendBlockingCommands() error {
...
 func (t *BlockingClientGroupTestCase) AssertResponses(logger *logger.Logger) error {
 	for _, clientWithExpectedResponse := range t.clientsWithExpectedResponses {
+		clientLogger := clientWithExpectedResponse.Client.GetLogger()
 		if clientWithExpectedResponse.Assertion == nil {
+			clientLogger.Infof("Expecting no response")
 			testCase := NoResponseTestCase{}
 			if err := testCase.Run(clientWithExpectedResponse.Client); err != nil {
 				return err
 			}
+			clientLogger.Successf("Success; no response received")
```
rohitpaulk, line 63: Can probably move this into `NoResponseTestCase` so it's symmetric with `ReceiveValueTestCase` which automatically logs a success message after completion.

Also, let's not use the "Success; ..." style - that's not something we do elsewhere. Our style would be something like "✔︎ No response received yet"

rohitpaulk (approved):

PR #198, opened 2025-07-14
Redis/Pub-Sub Extension

UdeshyaDhungana:
- Add extension in makefile
- Add extension in course definition
- remove bulk string length limitation in decoder
- Add subscribe response assertion
- Add error regex assertion

@rohitpaulk I had to make a small change in `internal/resp/decoder/decode_bulk_string_or_nil.go`. This case was missed in the stage breakdown document. 

The response to the `PING` when in subscribe mode is not `PONG`, but rather an array of two elements. The first element is "pong" (bulk string), and the second element is the argument supplied to the ping command. If there was no argument, the second element is a bulk string of length 0.

Screenshot 2025-07-15 at 13 02 43


internal/test_helpers/course_definition.yml
```diff
@@ -112,6 +112,17 @@ extensions:
       [rpush-command]: https://redis.io/docs/latest/commands/rpush/
       [lrange-command]: https://redis.io/docs/latest/commands/lrange/
 
+  - slug: "pub-sub"
+    name: "Pub-Sub"
```
rohitpaulk, line 116:
```suggestion
    name: "Pub/Sub"
```

(Consistent with Redis's official docs)

internal/resp/decoder/decode_bulk_string_or_nil.go
```diff
@@ -35,10 +35,8 @@ func decodeBulkStringOrNil(reader *bytes.Reader) (resp_value.Value, error) {
 		return resp_value.NewNilValue(), nil
 	}
 
-	if length < 1 {
-		// Ensure error points to the correct byte
+	if length < 0 {
 		reader.Seek(int64(offsetBeforeLength), io.SeekStart)
-
```
rohitpaulk, line 41: Would keep the newline here, consistent with the statement above

internal/resp_assertions/message_received_assertion.go
```diff
@@ -0,0 +1,22 @@
+package resp_assertions
+
+import (
+	resp_value "github.com/codecrafters-io/redis-tester/internal/resp/value"
+)
+
+type MessageReceivedAssertion struct {
+	channel string
```
rohitpaulk, line 8: Check other assertions and see how they name these - they're public (capitalized) + named like `ExpectedChannel`.

internal/resp_assertions/message_received_assertion.go
```diff
@@ -0,0 +1,22 @@
+package resp_assertions
+
+import (
+	resp_value "github.com/codecrafters-io/redis-tester/internal/resp/value"
+)
+
+type MessageReceivedAssertion struct {
```
rohitpaulk, line 7: The naming here is inconsistent with other assertions - review, if not sure share options and why you think they're similar/different and we can review

internal/resp_assertions/subscribe_response_assertion.go
```diff
@@ -0,0 +1,33 @@
+package resp_assertions
+
+import (
+	"fmt"
+
+	resp_value "github.com/codecrafters-io/redis-tester/internal/resp/value"
+)
+
+type SubscribeResponseAssertion struct {
+	channel         string
```
rohitpaulk, line 10: Same note as above, naming style is inconsistent with other assertions

internal/test_cases/pub_sub_test_case.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	// remove the channel from the subscriber only if it was previously present
+	if channelIndex != -1 {
+		t.subscribers[clientIndex].Channels = append(subscribedChannels[:channelIndex], subscribedChannels[channelIndex+1:]...)
+	}
+	return nil
+}
+
+func publish(t *PubSubTestCase, channel string, message string, publisher *instrumented_resp_connection.InstrumentedRespConnection, logger *logger.Logger) error {
```
rohitpaulk, line 98: This can be a method on the test case struct instead (`func (t *PubSubTestCase)`), very uncommon to define functions like this

internal/test_pubsub_publish1.go
```diff
@@ -0,0 +1,60 @@
...
+		AddSubscriber(clients[1], channels[1]).
+		AddSubscriber(clients[2], channels[1]).
+		SubscribeFromAll(logger)
+
+	if err != nil {
+		return err
+	}
+
+	publisherClient := clients[3]
+	publishTestCase := test_cases.MultiCommandTestCase{
```
rohitpaulk, line 47: Can't we use `pubSubTestCase` here? Surprised that we need to create these commands / assertions by hand when logic is already defined in that test case.

internal/test_pubsub_publish1.go
```diff
@@ -0,0 +1,60 @@
...
+		client-1 subscribes to channels[0]
+		client-2 and client-3 subscribe to channels[1]
+	*/
+
+	pubSubTestCase := test_cases.NewPubSubTestCase()
+
+	err = pubSubTestCase.
+		AddSubscriber(clients[0], channels[0]).
+		AddSubscriber(clients[1], channels[1]).
+		AddSubscriber(clients[2], channels[1]).
```
rohitpaulk, line 39: Would be cleaner to define the test case first (use `AddSubscriber` in the declaration above), and then do a single-line `if err = pubSubTestCase.SubscribeFromAll()` here

internal/test_pubsub_subscribe3.go
```diff
@@ -0,0 +1,86 @@
...
+		Assertion: resp_assertions.NewErrorRegexAssertion("^ERR (?i:Can't execute 'echo').*"),
+	}
+	if err := echoTestCase.Run(client, logger); err != nil {
+		return err
+	}
+
+	/* Test against PING and SUBSCRIBE (Allowed commands) */
+	pingTestCase := test_cases.SendCommandTestCase{
+		Command:   "PING",
+		Assertion: resp_assertions.NewOrderedStringArrayAssertion([]string{"pong", ""}),
```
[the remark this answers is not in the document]
rohitpaulk: I'd suggest making them do the disabling first, and then handle the ping case next

internal/test_cases/pub_sub_test_case.go
```diff
@@ -0,0 +1,143 @@
...
+type subscriber struct {
+	Client   *instrumented_resp_connection.InstrumentedRespConnection
+	Channels []string
+}
+
+type PubSubTestCase struct {
+	subscribers []subscriber
+}
+
+func NewPubSubTestCase() *PubSubTestCase {
```
rohitpaulk, line 20: The test case name itself is fine here, it's similar to `TransactionTestCase` for example. 

The methods could use some refinement though. "Subscriber" could commonly refer to single client, but the interface requires doing `AddSubscriber(client, channel1)` followed by `AddSubscriber(client, channel2)`. Maybe renaming this to `AddSubscription(client, channel)` would feel more natural? 

Most of our other test cases also use the `Run` terminology - like `RunMulti`, `RunWithoutExec` etc. Doing something similar here would make this more consistent. 

Let's take another look at the public interface here and try to make it align better with other similar ones. If you're not sure, share options and we can review

internal/test_pubsub_subscribe3.go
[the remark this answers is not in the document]
UdeshyaDhungana: Sure, I'll make a PR for this and course definition as well.

UdeshyaDhungana:
Applied requested changes:

- Rename Pub-Sub -> Pub/Sub
- Revert to keeping newline in `decode_bulk_string_or_nil.go`
- Rename MessageReceivedAssertion -> PublishedMessageAssertion
- Rename struct fields to Expected... in newly added assertions
- Rename all methods to RegexErrorAssertion
- Added getSubscriberFromClient method and used that in `RunUnsubscribe`
- Used `pubSubTestCase` in `test_pubsub_publish1.go`
- Rename `AddSubscriber`- > `AddSubscription`
- Split subscribe3 -> disallow commands and modify ping stages
- Tried renaming methods in `pub_sub_test_case.go`

internal/test_cases/pub_sub_test_case.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+		if err := subscribeTestCase.RunAll(subscriber.Client, logger); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (t *PubSubTestCase) getSubscriberFromClient(client *instrumented_resp_connection.InstrumentedRespConnection) (subscriber subscriber, idx int) {
```
rohitpaulk, line 57: To stay consistent with other files, keep all public methods first and then the private methods below them. This allows quick scanning to figure out what the public interface for a package is.

internal/test_cases/pub_sub_test_case.go
rohitpaulk: This is now better than before, but still not really simple. The current interface is: 

```
- AddSubscription(client, channel)
- RunSubscribeFromAll(logger)
- RunUnsubscribe(client, channel)
- RunPublish(channel, message, publisherClient, logger)
- RunPublishWithoutMessageAssertion(channel, message, publisherClient, logger)
```

I think this has the same issue as what `BlockingClientGroupTestCase` previously had, it's trying to do too much. The only test case I can think of with a larger interface is `TransactionTestCase`, but there it kinda makes sense because `RunAll()` is the primary method and everything else is just splitting out portions of `RunAll()` to be called independently. 

Thoughts on splitting this? What if we did `SubscriberGroupTestCase` and `PublishTestCase` separately? 

```
type SubscriberGroupTestCase {}

- AddSubscription(client, channel)
- RemoveSubscription(client, channel)
- GetSubscriberCount(channel)

- RunSubscribe(logger)
- RunAssertionsForPublishedMessage(channel, message, logger)
```

```
type PublishTestCase {
  Channel string
  Message string
  ExpectedSubscriberCount int
}

- Run(client, logger)
```

```
type UnsubscribeTestCase {
  Channel string
  Message string
  ExpectedSubscriberCountAfterUnsubscribe int
}

- Run(client, logger)
```

rohitpaulk (changes requested):
Added one more round of comments

PR #200, opened 2025-07-21
Add NewSimpleStringAssertion for PING/PONG

andy1li:
Context:

https://forum.codecrafters.io/t/update-tests-to-be-more-specific-simple-string-vs-bulk-string/13707

https://redis.io/docs/latest/commands/ping/#return-information

image


rohitpaulk:
(Expect a few people to complain that this broke their implementation - that's okay though)

PR #202, opened 2025-07-22
fix: simple resp error value format

UdeshyaDhungana:
Previous log message in case of RESP simple error
Screenshot 2025-07-22 at 08 59 20

Newly formatted log message:
Screenshot 2025-07-22 at 08 58 59

The `ERR` is not a part of resp simple error (Ref: https://redis.io/docs/latest/develop/reference/protocol-spec/#simple-errors). So, let's remove it from the log messages.

## Summary by CodeRabbit

* **Bug Fixes**
  * Error messages no longer include redundant "ERR:" prefixes, resulting in clearer and more consistent error output.

* **Tests**
  * Updated test output logs to reflect corrected error message formatting, new replication IDs, timestamps, port numbers, and other environment-generated values.
  * Adjusted expected outputs for command timings, key order, stream entry IDs, and file paths to match current results.
  * No changes to test logic or behavior.

### DIFF
--- a/internal/resp/value/value.go
+++ b/internal/resp/value/value.go
@@ -118,7 +118,7 @@ func (v *Value) FormattedString() string {
 		}
 		return formatter.Prettify(respJson)
 	case ERROR:
-		return fmt.Sprintf("%q", v.String())
+		return fmt.Sprintf("%q (error)", v.String())
 	case NIL:
 		return "\"$-1\\r\\n\""
 	}
--- a/internal/resp/value/value.go
+++ b/internal/resp/value/value.go
@@ -118,7 +118,7 @@ func (v *Value) FormattedString() string {
 		}
 		return formatter.Prettify(respJson)
 	case ERROR:
-		return fmt.Sprintf("%q", "ERR: "+v.String())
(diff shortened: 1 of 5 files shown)


internal/resp/value/value.go
```diff
@@ -118,7 +118,7 @@ func (v *Value) FormattedString() string {
 		}
 		return formatter.Prettify(respJson)
 	case ERROR:
-		return fmt.Sprintf("%q", "ERR: "+v.String())
```
rohitpaulk, line 121: I do think it's useful to have some indication that this is an error type - like how an integer is `123` and a string is `"123"`. How about `"<error message>" (error)`?

rohitpaulk (changes requested):
Q added

internal/resp/value/value.go
rohitpaulk: Okay can't think of anything at the moment - @UdeshyaDhungana let's go with your version that doesn't have the duplication! Can think of a more proper solution later

rohitpaulk (approved):

PR #203, opened 2025-07-25
Patch: redis-cli logging

UdeshyaDhungana:
- In pub sub test cases, use shorter messages
- Re-format `redis-cli` logging.

## Summary by CodeRabbit

* **New Features**
  * Improved logging of command-line arguments with enhanced formatting for better readability.

* **Bug Fixes**
  * Updated log message formatting in test cases to use quoted strings for clearer output.

* **Tests**
  * Test messages now use random words instead of arbitrary strings, improving test clarity and relevance.

### DIFF
--- a/internal/instrumented_resp_connection/utils.go
+++ b/internal/instrumented_resp_connection/utils.go
@@ -10,7 +10,7 @@ func quoteIfHasSpaceOrEscapeSequence(s string) string {
 	quoted := fmt.Sprintf("%q", s)
 
 	// if the string does not change on trimming double quotes, no escapable chracter was present
-	if s == strings.Trim(quoted, "\"") && !strings.Contains(s, " ") {
+	if (s == strings.Trim(quoted, "\"")) && (!strings.Contains(s, " ")) {
 		return s
 	}
 


internal/instrumented_resp_connection/utils.go
```diff
@@ -0,0 +1,28 @@
...
+	// if the string does not change, no escapable chracter was present
+	if s == trimmedQuotes && !strings.Contains(s, " ") {
+		return s
+	}
+
+	return quoted
+}
+
+// quoteCLIArgs applies quoteIfNeeded to each CLI argument
+func quoteCLIArgs(args []string) []string {
```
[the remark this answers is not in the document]
rohitpaulk: Ah no, this is confusing for sure - let's stick to our version then!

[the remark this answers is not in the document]
UdeshyaDhungana: Sure!

internal/instrumented_resp_connection/utils.go
```diff
@@ -0,0 +1,28 @@
...
+	// if the string does not change, no escapable chracter was present
+	if s == trimmedQuotes && !strings.Contains(s, " ") {
+		return s
+	}
+
+	return quoted
+}
+
+// quoteCLICommand applies quoteIfHasSpaceOrEscapeSequence to each item in commandWithArgs
+func quoteCLICommand(commandWithArgs []string) []string {
```
rohitpaulk, line 22:
```suggestion
func quoteCLICommand(commandWithArgs []string) string {
```

Let's make this just return a single string - similar to the library we looked at earlier. If we're quoting args, it's very unlikely we need them individually. Quoting is almost always done when you need to pass to a shell or log like we're doing - in both of those cases, the input will be concatenated before use.

internal/instrumented_resp_connection/utils.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// quoteIfHasSpaceOrEscapeSequence quotes a string if it contains escapable characters or spaces
+func quoteIfHasSpaceOrEscapeSequence(s string) string {
+	quoted := fmt.Sprintf("%q", s)
+	trimmedQuotes := strings.Trim(quoted, "\"")
```
rohitpaulk, line 11: Bit of a weird name - `trimmedQuotes` doesn't make it obvious what the variable contains. @UdeshyaDhungana want to come up with other options here (or just avoid the intermediary variable altogether?)?

internal/instrumented_resp_connection/utils.go
UdeshyaDhungana: Got it!

PR #204, opened 2025-08-05
Sorted Sets Extension

UdeshyaDhungana:


internal/resp_assertions/floating_point_bulk_string_assertion.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func (a FloatingPointBulkStringAssertion) Run(value resp_value.Value) error {
+	if value.Type != resp_value.BULK_STRING {
+		return fmt.Errorf("Expected bulk string, got %s", value.Type)
+	}
+
+	stringValue := value.String()
+	f, err := strconv.ParseFloat(stringValue, 64)
```
rohitpaulk, line 28:
```suggestion
	floatValue, err := strconv.ParseFloat(stringValue, 64)
```

Avoid single-letter variable names in general, only exception is struct method target (like `a` here), or `i`/`j` in a loop

internal/resp_assertions/floating_point_bulk_string_assertion.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		return fmt.Errorf("Expected %q to be a floating point number", stringValue)
+	}
+
+	diff := f - a.ExpectedValue
```
rohitpaulk, line 33: Can use `Math.abs` here

rohitpaulk (approved):

PR #209, opened 2025-08-15
Redis Geospatial Stages

UdeshyaDhungana:


internal/data_structures/location.go
```diff
@@ -0,0 +1,176 @@
...
+
+type Coordinates struct {
+	Latitude  float64
+	Longitude float64
+}
+
+// Location : I'll remove this comment later
+// I'm not sure if making a strucutre like `LocationSet` is a good idea
+// Given that the data will not encapsulate any new information regarding the locations (unlike ZSET where order, etc is preserved)
+// I didn't find it necessary to create one
```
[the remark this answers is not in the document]
rohitpaulk: I'm not sure if the `{}Location -> {}string[]` conversion is something you'd want to handle on `LocationSet` though - depends on context (haven't seen enough usages to know)

internal/test_cases/geopos_test_case.go
```diff
@@ -0,0 +1,43 @@
...
+type GeoPosTestCase struct {
+	Key                 string
+	LocationNames       []string
+	Tolerance           float64
+	ExpectedCoordinates []*data_structures.Coordinates
+}
+
+// Run : I'll remove this comment later
+// I wonder if this can be improved, stuck in a dilemma.
+// From its use in `test_geospatial_geopos.go`, it is not apparent that nil is converted to NilAssertion()
```
rohitpaulk, line 19: Good question.. How about accepting `Locations []Location`, and `MissingLocations []Location`? Then within the test case we could construct all the assertions required?

We can remove `Tolerance` here too, don't think we ever customize it.

rohitpaulk (changes requested):
Comments added! Only focused on the questions you had for now - not looking at other parts

internal/test_cases/geoadd_test_case.go
```diff
@@ -0,0 +1,27 @@
...
+type GeoAddTestCase struct {
+	Key                         string
+	Location                    data_structures.Location
+	ExpectedAddedLocationsCount int
+}
+
+func (t *GeoAddTestCase) Run(client *instrumented_resp_connection.InstrumentedRespConnection, logger *logger.Logger) error {
+	longitudeStr := strconv.FormatFloat(t.Location.GetLongitude(), 'f', -1, 64)
+	latitudeStr := strconv.FormatFloat(t.Location.GetLatitude(), 'f', -1, 64)
+	sendCommandTestCase := SendCommandTestCase{
```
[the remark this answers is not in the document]
rohitpaulk: @UdeshyaDhungana think this is still pending

internal/test_cases/geosearch_test_case.go
```diff
@@ -0,0 +1,32 @@
...
+	"strconv"
+
+	"github.com/codecrafters-io/redis-tester/internal/instrumented_resp_connection"
+	"github.com/codecrafters-io/redis-tester/internal/resp_assertions"
+	"github.com/codecrafters-io/tester-utils/logger"
+)
+
+type GeoSearchTestCase struct {
+	Key                   string
+	FromLongitude         float64
```
rohitpaulk, line 13: Accept `FromCoordinates` instead. That lets us use the `AsRedisCommandArg()`... function below.

internal/test_cases/geoadd_test_case.go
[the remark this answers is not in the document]
rohitpaulk: This file looks correct now - will try to come up with some rules to codify this!

internal/test_cases/geopos_test_case.go
```diff
@@ -0,0 +1,69 @@
...
+type GeoPosTestCase struct {
+	Key                  string
+	Locations            []location.Location
+	MissingLocationNames []string
+	// If ShouldVerifyCoordinates is true, only floating point parsing is checked for existing locations
+	ShouldVerifyCoordinates bool
+}
+
+func (t *GeoPosTestCase) Run(client *instrumented_resp_connection.InstrumentedRespConnection, logger *logger.Logger) error {
+
```
rohitpaulk, line 22: Extra newline at start of function

internal/test_cases/geopos_test_case.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/codecrafters-io/tester-utils/logger"
+	"github.com/codecrafters-io/tester-utils/random"
+)
+
+type GeoPosTestCase struct {
+	Key                  string
+	Locations            []location.Location
+	MissingLocationNames []string
+	// If ShouldVerifyCoordinates is true, only floating point parsing is checked for existing locations
+	ShouldVerifyCoordinates bool
```
rohitpaulk, line 18:
```suggestion
	ShouldSkipCoordinateVerification bool
```

or `ShouldSkipVerifyingCoordinates`, `ShouldOnlyVerifyCoordinatesAreFloats` etc.

(The obvious case for `GeoPosTestCase` is to verify coordinates - the skipping one seems like a edge-case, would not make it the default)

internal/test_cases/geopos_test_case.go
```diff
@@ -0,0 +1,69 @@
...
+	for i, missingLocationName := range t.MissingLocationNames {
+		allLocationNames[offset+i] = missingLocationName
+		// WILL_CHANGE: NewNilArrayAssertion() after PR#211 is merged
+		allAssertions[offset+i] = resp_assertions.NewNilAssertion()
+	}
+
+	// Shuffle location names and assertions in same order
+	shuffledLocationNames := make([]string, allLocationsLen)
+	shuffledAssertions := make([]resp_assertions.RESPAssertion, allLocationsLen)
+	shuffledIndexes := random.RandomInts(0, allLocationsLen, allLocationsLen)
```
rohitpaulk, line 55: We've got `ShuffleArray` inside tester-utils, and I think it accepts a generic parameter (let's change it if not)

internal/test_cases/geopos_test_case.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for i, idx := range shuffledIndexes {
+		shuffledLocationNames[i] = allLocationNames[idx]
+		shuffledAssertions[i] = allAssertions[idx]
+	}
+
+	sendCommandTestCase := SendCommandTestCase{
+		Command:   "GEOPOS",
+		Args:      append([]string{t.Key}, shuffledLocationNames...),
```
rohitpaulk, line 64: Are we only running one of these tests? I thought that we were running multiple (the instructions say so)

rohitpaulk: At the very least we should test: (a) only missing location (b) only valid location and (c) mix of multiple valid & missing locations

internal/test_geospatial_calculate_score.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+	defer client.Close()
+
+	locationKey := testerutils_random.RandomWord()
+	locationSet := location.GenerateRandomLocationSet(testerutils_random.RandomInt(2, 4))
+	locations := locationSet.GetLocations()
+
+	// Add locations
+	for _, loc := range locations {
```
rohitpaulk, line 33:
```suggestion
	for _, location := range locations {
```

(Avoid abbreviations unless it's a struct method receiver)

internal/test_geospatial_geodist.go
```diff
@@ -0,0 +1,77 @@
...
+			if err := geodistTestCase.Run(client, logger); err != nil {
+				return err
+			}
+		}
+	}
+
+	// Test distance from a loc to itself (should be 0)
+	loc := testerutils_random.RandomElementFromArray(locations)
+
+	if len(locations) > 0 {
```
rohitpaulk, line 63: Unsure what this check is for - under what circumstances would `locations` be an empty array?

internal/test_geospatial_geopos.go
```diff
@@ -0,0 +1,73 @@
...
+
+		if err := geoAddTestCase.Run(client, logger); err != nil {
+			return err
+		}
+	}
+
+	missingLocationNames := make([]string, random.RandomInt(2, 4))
+
+	for i := range len(missingLocationNames) {
+		missingLocationNames[i] = fmt.Sprintf("missing_location_%d", random.RandomInt(1, 100))
```
rohitpaulk, line 49: It's possible to have duplicates here if we call `random.RandomInt` - not a problem to test this case but it might be weird in logs to see `missing_location_1 missing_location_1`, could look like a bug. I'd use `RandomInts(1, 100, len(...))` instead

rohitpaulk (approved):
Minor comments - good to merge once those are resolved!

PR #212, opened 2025-08-20
Add docker file for recording fixtures and running tests

UdeshyaDhungana:
- Stages other than `yg4` (previously failing tests due to extra logs in fixtures) do not have command propagation tests. 
- Added a small delay to get rid of extra log in fixtures

### DIFF
--- a/docker_ci/ubuntu.Dockerfile
+++ b/docker_ci/ubuntu.Dockerfile
@@ -1,11 +1,7 @@
-# Use Ubuntu latest as base image
-FROM ubuntu:latest
+# Use official Go 1.24 image as base
+FROM golang:1.24-bookworm
 
 # Set environment variables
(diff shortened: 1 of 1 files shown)


docker_ci/ubuntu.Dockerfile
```diff
@@ -0,0 +1,60 @@
+# Use Ubuntu latest as base image
```
rohitpaulk, line 1: Can we just use Go as the base? Would make this file far shorter

docker_ci/docker-ci.sh
rohitpaulk, line 1: This file is fine to keep for now - just adding a note that we should eventually standardize something like this across all testers so that we don't have to duplicate this for every tester. Looks like we're doing this in grep-tester, here and in shell-tester at the moment.

rohitpaulk (approved):

PR #215, opened 2025-08-25
Transform callback: Approach 1

UdeshyaDhungana:


rohitpaulk:
So it's a bit tricky to choose between these two - the pros to this approach is that the changes here are more contained compared to Approach 2 and thus the diff is cleaner to read. 

The downsides are (a) the interface for `TransformReceivedBytes` is more complicated now (and getting the decodedLength wrong could result in weird errors) (b) we do change `c.ReceivedBytes` (the count of received bytes) which is kinda inaccurate. If we were using this in one of the replication stages for example, the ACKs would be incorrect because they'd be using a length from the transformed value instead of the original received bytes count.

Just noting this down, thinking about what to go with...

rohitpaulk:
Let's go with approach 2 to avoid the two downsides mentioned above - we can structure our naming there better to make the behaviour a little less confusing though. Will leave comments on the other PR

PR #216, opened 2025-08-25
Transform callback: Approach 2

UdeshyaDhungana:


UdeshyaDhungana:
In this approach, we leave the `UnreadBuffer` property untouched, and use the transformed value for logging and converting to resp value.

internal/instrumented_resp_connection/instrumented_resp_connection.go
```diff
@@ -85,3 +85,15 @@ func (c *InstrumentedRespConnection) UpdateBaseLogger(l *logger.Logger) {
 	c.logger = newLogger
 	c.UpdateCallBacks(defaultCallbacks(c.logger))
 }
+
+// SetReceivedBytesTransform sets the TransformReceivedBytes parameter of the instrumented resp connection
```
rohitpaulk, line 89: Would remove these two comments, don't seem very helpful

internal/resp/connection/connection.go
```diff
@@ -145,14 +149,22 @@ func (c *RespConnection) ReadFullResyncRDBFile() ([]byte, error) {
...
+		readBytesCount = c.UnreadBuffer.Len()
 	}
-	if c.Callbacks.AfterBytesReceived != nil && loggableBytesCount > 0 {
-		c.Callbacks.AfterBytesReceived(c.UnreadBuffer.Bytes()[:loggableBytesCount])
+
+	readBytes := c.UnreadBuffer.Bytes()[:readBytesCount]
+
+	if c.Callbacks.TransformReceivedBytes != nil {
+		readBytes = c.Callbacks.TransformReceivedBytes(readBytes)
+		value, _, _ = resp_decoder.DecodeFullResyncRDBFile(readBytes)
```
rohitpaulk, line 163: We shouldn't be ignoring these errors, it's a great chance to ensure we don't have subtle bugs - if err is not nil here, this should be an internal error. Similarly if the decoded length doesn't match `readBytes`, that should be an internal error too

internal/resp/connection/connection.go
```diff
@@ -145,14 +149,22 @@ func (c *RespConnection) ReadFullResyncRDBFile() ([]byte, error) {
...
-	loggableBytesCount := readBytesCount
+	readBytesCount := decodedBytesCount
 	if err != nil {
-		loggableBytesCount = c.UnreadBuffer.Len()
+		readBytesCount = c.UnreadBuffer.Len()
 	}
-	if c.Callbacks.AfterBytesReceived != nil && loggableBytesCount > 0 {
-		c.Callbacks.AfterBytesReceived(c.UnreadBuffer.Bytes()[:loggableBytesCount])
+
+	readBytes := c.UnreadBuffer.Bytes()[:readBytesCount]
```
rohitpaulk, line 159: The naming/structure here needs some work overall - the main part that I think is unclear: `readBytesCount` can be different from `len(readBytes)` after the transformation has occurred.

rohitpaulk (changes requested):

PR #219, opened 2025-08-27
Improve #ec3 Blocking retrieval 

andy1li:
1. Send commands from clients in the order of increasing number

2. Assert the `NoExpectedResponse` first

image


internal/test_cases/blocking_client_group_test_case.go
```diff
@@ -56,20 +56,37 @@ func (t *BlockingClientGroupTestCase) SendBlockingCommands() error {
...
-			if err := testCase.Run(clientWithExpectedResponse.Client, logger); err != nil {
-				return err
-			}
+		if err := t.assertResponse(&clientWithExpectedResponse, logger); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
```
rohitpaulk, line 65: Do we really need two versions? Could we get away with just `AssertResponses` and have that run in reverse order?

rohitpaulk (changes requested):
Added a q

internal/test_cases/blocking_client_group_test_case.go
```diff
@@ -18,7 +18,8 @@ type clientWithExpectedResponse struct {
 }
 
 type BlockingClientGroupTestCase struct {
-	clientsWithExpectedResponses []clientWithExpectedResponse
+	clientsWithExpectedResponses        []clientWithExpectedResponse
+	ShouldAssertResponsesInReverseOrder bool
```
rohitpaulk, line 22: @andy1li my question was more related to do we need to ever control the order of assertions? Can we always use the reverse order? (Does the logic not apply to all stages?)

rohitpaulk (changes requested):
Added q

internal/test_cases/blocking_client_group_test_case.go
rohitpaulk: Yep! Can remove the configuration option and keep things simple

PR #222, opened 2025-09-18
Blocking retrieval non ordering

UdeshyaDhungana:


rohitpaulk:
Non-trivial change so want to take a closer look - will look at tomorrow!

